PR #5215, opened 2014-04-10
nodejs cartridge: use npm install --production

Filirom1:
In the [npm config doc](https://www.npmjs.org/doc/misc/npm-config.html)  it's explained that `-d` means `--loglevel info`.

Moreover,

> By default, npm install will install all modules listed as dependencies. With the --production flag, npm will not install modules listed in devDependencies.

https://www.npmjs.org/doc/cli/npm-install.html

`devDependencies` are used for tests. We don't want to download tests dependencies when we deploy to openshift. It will make build faster for nodejs projects.


marvinroger:
Is there anything new? Applications these days have so much devDeps, with gulp, mocha, browserify, stylus, all this kind of stuff. Deploying litterally takes tens of minutes!

PR #5314, opened 2014-04-21
correct deprecated node.js cartridge

nightmared:
The code used in previous version of the node.js cartridge was deprecated and this minor commit correct the problem. The express.createServer() was deprecated and it's better to use the new recommend method instead of using a function that could disappears at any update of the express framework.


danmcp:

PR #5506, opened 2014-06-13
OpenJDK 8 support for jbossews and jbosseap

jerboaa:
Hi,

Here is a pull request which adds support for OpenJDK 8 for jbossews and jbosseap cartridges. It has been done in a similar way than OpenJDK 7 support is done. I.e. app users create a marker file called java8 which makes the services use OpenJDK 8 for the hosting JVM.

I've added cartridge lifecycle tests for both.

Testing done:
$ cucumber openshift-test/tests/cartridge-lifecycle-jbosseap.feature
$ cucumber openshift-test/tests/cartridge-lifecycle-jbossews.feature

Both pass for me.

Note that a java-1.8.0-openjdk package needs to be available for things to install.

Thoughts?


danmcp:

PR #5806, opened 2014-09-11
Add oo-admin-geartop

sdodson:
oo-admin-geartop provides a `top` like summary of per gear CPU and Memory consumption with an understanding of cgroups resource limits.

Example output :

```
            GEARTOP -- Update: 3s  Sort: cpu_pct  09/11/14 14:16:43             
                 uuid              mem_MB   swap_MB  cpu_total cpu_quota% throttled cpu%
         5411e6a83a0fb2d397000014    70.51     0.00      45.24    100.0       5.49 104.15
         540f72243a0fb2f854000001    16.04     0.00      15.39    100.0       0.05  0.00
-- displaying 2 gears. Press CTRL-c twice to quit.
```


twiest:
@sdodson @jwhonce @kwoodson   We had a customer call today, and they were asking for this type of functionality.

Is this something we can get cleaned up and merged?

node-util/sbin/oo-admin-geartop
```diff
@@ -0,0 +1,188 @@
...
+        uuids = get_usernames(uids)
+
+        data = refresh_gear_data(uuids, previous, interval)
+        limit = [ uuids.size, %x[ tput lines ].to_i - 4 ].min
+
+        sort_value = sort['c'] if sort_value.nil?
+        strings = []
+        if cnt > 0
+          strings += data.sort_by do |k,v|
+              v[sort_value.to_sym]
```
sdodson: I can't get this to leak, i think given we're operating on a predefined hash of 5 items we should be safe.

node-util/sbin/oo-admin-geartop
```diff
@@ -0,0 +1,188 @@
...
+          end.reverse.first(limit).collect { |gear| gear[1] }
+        else
+          strings += data.values.collect { |gear| gear }
+          next
+        end
+
+        system("clear")
```
sdodson: I did a bit of searching for other methods to clear the terminal and as far as I can tell this is the most portable method for clearing the terminal. I tried %x[clear] and that failed, I tried sending a predefined escape sequence and that works great for xterm / gnome-terminal but it sounds like it doesn't work well for mac terminals. clear, which is part of ncurses examines the termcap and picks the appropriate escape sequence to clear the terminal.

sdodson:
@jwhonce do you mind merging this as is?

jwhonce:
[merge]

PR #5853, opened 2014-10-06
Add option to have the default application URL use https

timothyh:
BZ #1144894 [RFE] Use https in application URL's


Miciah:
`APP_ADVERTISE_HTTPS` would be more apt a name for the setting seeing that we allow both in any case.

danmcp:

PR #5959, opened 2014-11-14
Bug 1164119: Show 10gen agent as having dependencies

liggitt:


danmcp:
@liggitt Was this waiting on something?

liggitt:
@danmcp someone who knows whether the cartridge was working well to decide whether we wanted this cartridge to be more visible

PR #5986, opened 2014-11-27
Bug 1165719 - oo-accept-node incorrectly using manifests from cart repository

detiber:
- oo-accept-node should use the manifest from the installed cartridge
  instead of from the cartridge repository, otherwise downloadable
  carts are not verified.


danmcp:
@bparees Review please.

bparees:
the change looks ok, but do we really want this in before stagecut this release?

bparees:
@danmcp @detiber @maxamillion ^^^

danmcp:
@bparees I wasn't pushing to get it before stage cut.  I was just going through the origin-server pulls in general.

danmcp:
[merge]

PR #6008, opened 2014-12-10
BUG 1171066: disable_auto_scaling marker requires restart of haproxy cartridge

jhadvig:
Updating the documentation for the HAProxy disable_auto_scaling marker, cause the cartridge needs to be restarted after the marker is added + adding an banner that warns user of the presence of the marker


jhadvig:
[test]

mfojtik:
LGTM [merge]

PR #6010, opened 2014-12-10
Bug 1171289 - background pkill command in OOMPlugin

a13m:
There are two significant fixes here:

1) Instead of using ApplicationContainer.kill_procs, which waits for pkill/pgrep process to return, we simply spawn a pkill and detach it, because we don't actually care about its return code, and we cannot wait for it.

2) The while loop runs when oom_control['under_oom'] is 1, even if the memory usage indicates that the cgroup has fallen below its memsw limit.


danmcp:
@a13m Looks like test_oom_control needs a corresponding change

jwhonce:
[merge]

PR #6030, opened 2014-12-29
Bug1176491: pass the pear channel-update with error due to the internet access problem

nak3:
See also [bz1176491](https://bugzilla.redhat.com/show_bug.cgi?id=1176491)

The root cause of bz1176491 is that the script bash [sets "-e" option](https://github.com/openshift/origin-server/blob/master/cartridges/openshift-origin-cartridge-php/bin/install#L1).
So, putting "&& true" like below will be able to fix it.

I know it is better if showing warning messages if the channel-update didn't work out. However, to output such a message needs several change in the other components more than openshift-origin-cartridge-php/bin/install.

So I think putting "&& true" is OK for this issue.


sosiouxme:
[merge]

PR #6033, opened 2015-01-05
broker: mark broker httpd confs as conf files

sosiouxme:
https://bugzilla.redhat.com/show_bug.cgi?id=1163917
Bug 1163917 - /var/www/openshift/broker/httpd/broker.conf is not marked
as a configuration file


sosiouxme:
[merge]

PR #6036, opened 2015-01-08
Bug 1179923: Excluding Perl pragmas added by perl.req

jhadvig:
Based on the bug I've created an array of well-known perl pragmas which can be excluded from installing dependencies in the build process, to avoid conflicts with cpan

### DIFF
--- a/cartridges/openshift-origin-cartridge-perl/bin/build
+++ b/cartridges/openshift-origin-cartridge-perl/bin/build
@@ -13,6 +13,7 @@ fi
 LINUX_DISTRO=$(</etc/redhat-release)
 RED_HAT_DISTRO_NAME="Red Hat"
 MIRROR="--mirror http://search.cpan.org/CPAN"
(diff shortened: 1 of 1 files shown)


jhadvig:
[test][extended:cartridge]

jhadvig:
@mfojtik could you please review, sice you are also pretty familiar with Perl cart.
Thx

mfojtik:
[merge]

cartridges/openshift-origin-cartridge-perl/bin/build
```diff
@@ -44,6 +45,11 @@ then
 
   for f in $(( echo "$DEPS" | xargs /usr/lib/rpm/perl.req | awk '{ print $1 }' | sed 's/^perl(\(.*\))$/\1/'; cat ${CPAN_PATH} 2>/dev/null ) | sort -u | grep . )
   do
+    # Skipping well-known perl pragmas
+    for element in ${EXCLUDED_PRAGMAS[@]}; do
+      [[ $element == $f ]] && continue
```
sosiouxme: I don't think "continue" here will break out of the loop you intended... this is basically a no-op.

PR #6039, opened 2015-01-09
node-msg: rm duplicate+inaccurate agent validations

sosiouxme:
MCollective already validates all request inputs via openshift.ddl, so
there is no need for the agent code to repeat validations. Further, the
validations no longer match what's in openshift.ddl. This led to some
bugs with the recent expansion of gear uuid format.

Bug 1176065 - Failed upgrade when gear_uuid is in predictable gear uuid format
https://bugzilla.redhat.com/show_bug.cgi?id=1176065

Bug 1180004 - Moving gear from non-district node to district node will
fail when set USE_PREDICTABLE_GEAR_UUIDS=true
https://bugzilla.redhat.com/show_bug.cgi?id=1180004

### DIFF
--- a/plugins/msg-node/mcollective/src/openshift.rb
+++ b/plugins/msg-node/mcollective/src/openshift.rb
@@ -113,7 +113,6 @@ module MCollective
 
 
       def echo_action
-        validate :msg, String
         reply[:msg] = request[:msg]
       end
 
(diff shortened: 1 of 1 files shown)


sosiouxme:
@jwhonce or @pmorie - are these agent validations actually necessary for some use case or just inaccurate duplication of what's already validated via openshift.ddl?

sosiouxme:
[test][extended:node]

sosiouxme:
honce gave thumbs up in irc, so [merge]

PR #6041, opened 2015-01-13
Bug 1179923: Fix update

jhadvig:
Just a small update so the Perl pragmas will be skipped in the installation loop


mfojtik:
[merge]

PR #6050, opened 2015-01-16
Use EXTERNAL_ETH_DEV to determine the node IP

codificat:
Facter was using PUBLIC_NIC to determine host_ip, but node.conf specifies EXTERNAL_ETH_DEV for the purpose of identifying the NIC for public traffic.

I haven't found other references to PUBLIC_NIC elsewhere; also I believe INTERNAL_ETH_DEV is not used anywhere, so changed facter to use EXTERNAL_ETH_DEV to determine host_ip.

After that, use that value instead of hardcoding eth0 as the source of the node's IP address.

### DIFF
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -690,24 +690,26 @@ module OpenShift
     end
 
     # <<accessor>>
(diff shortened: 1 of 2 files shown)


dobbymoodge:
+1 LGTM - `scl enable ruby193 'facter host_ip'` gives the expected value, and correctly uses `EXTERNAL_ETH_DEV` when specified.

brenton:
[merge]

PR #6052, opened 2015-01-16
Regions

kwoodson:
Updated admin_stats_test and friends to work with new region support.  Unit tests were failing so I have added the necessary fixes for these tests to pass.   This includes adding 2 new additional tests that test getting a region and getting a region by passing a district.  The summary portion is left untested for now as it would require much more node info in order to test.

### DIFF
--- a/controller/app/models/admin/stats/maker.rb
+++ b/controller/app/models/admin/stats/maker.rb
@@ -186,7 +186,7 @@ class Admin::Stats::Maker
     #   "6e5d3ccc0bb1456399687c0be51676f8" => ...
     # }
 
-    fields = %w[uuid name gear_size servers max_capacity available_capacity available_uids servers]
+    fields = %w[uuid name gear_size servers max_capacity available_capacity available_uids]
     with_each_record(:districts, {}, {:fields => fields }) do |dist|
       entry_for_district[dist['uuid'].to_s] = Admin::Stats::DistrictEntry.new.merge({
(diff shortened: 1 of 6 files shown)


kwoodson:
[test]

kwoodson:
@abhgupta  please review this commit as we would like to get this into our stage/production environments.

sosiouxme:
I suspect @abhgupta will be out for a bit. I can take a look at it. @sosiouxme poke to self

kwoodson:
[test]

controller/app/models/admin/stats/maker.rb
```diff
@@ -179,7 +186,7 @@ def get_district_entries
     #   "6e5d3ccc0bb1456399687c0be51676f8" => ...
     # }
 
-    fields = %w[uuid name gear_size servers max_capacity available_capacity available_uids]
+    fields = %w[uuid name gear_size servers max_capacity available_capacity available_uids servers]
```
sosiouxme: Don't think it makes much difference, but "servers" is duplicate here.

kwoodson:
[test]

sosiouxme:
@kwoodson I did some tests with a 6-node, 2-profile deployment, with and without regions. I checked the admin-console didn't bork (it doesn't display the new data, of course). Looked through the code a bit. Looks good to me. [merge] per @danmcp

PR #6057, opened 2015-01-26
tuned-profiles-openshift-node: Don't load /etc/sysctl.ktune

sdodson:
RHEL6 tuned no longer uses this file nor do we ship it, so by default don't
attempt to load it. This only affects things if a user had created
/etc/sysctl.ktune on their own which is not likely.


brenton:
LGTM.  Can you merge this yourself or do you need me to submit it?

sdodson:
[merge]

sdodson:
@brenton looks like I need you to merge it

brenton:
[merge]

PR #6060, opened 2015-01-27
BZ1186171 - scp needs additional arguments to avoid prompts

calfonso:
- Improper scp command cause the failure of copying certificate and key
  for alias to BIGIP LTM host


calfonso:
[merge]

PR #6061, opened 2015-01-30
BZ1187047 - "oo-admin-ctl-routing delete-alias" does not delete..

calfonso:
- "oo-admin-ctl-routing delete-alias" does not delete related alias cert
  files.


calfonso:
[merge]

PR #6065, opened 2015-02-02
Protect 'gear' command from gems in $GEM_HOME

ncdc:
Explicitly set GEM_HOME to '' to protect the 'gear' command and the ruby
SDK from running with user-supplied gems.

Bug 1187829


ncdc:
[test][extended:gear,node]

ncdc:
[merge]

PR #6069, opened 2015-02-10
Bug 1191238 - Allow domain to be specified in oo-admin-ctl-app

tiwillia:
Bugzilla Link https://bugzilla.redhat.com/show_bug.cgi?id=1191238
Allow domain to be specified in oo-admin-ctl-app


brenton:
[merge]

PR #6070, opened 2015-02-11
Feature/submodule clone deploybranch

sedninja:
This pull request addresses an issue that can arise when deploying an application with submodules using a branch other than `master`.


danmcp:
@sedninja This would imply that the main git repo and the submodule must be working off the same branch?  That seems just as arbitrary.

danmcp:

PR #6071, opened 2015-02-12
Handle all encodings during regex comparisons

ironcladlou:
Open files in binary mode to support regex matching on all encodings.

Fix bug #1191681


ironcladlou:
[test]

PR #6073, opened 2015-02-12
Handle all encodings during regex comparisons

ironcladlou:
Open files in binary mode to support regex matching on all encodings.

Fix bug #1191681


ironcladlou:
[test]

ironcladlou:
@jwhonce please review and tag if it looks okay. In addition to the expanded func test, I also did manual tests with this approach in irb and I'm pretty sure it's good.

jwhonce:
[merge]

PR #6075, opened 2015-02-16
Bug 1191181 - Added checking server certs existence when turning on SSL.

soltysh:
For SSL in postgresql to work properly it's needed to have `server.crt` and `server.key` defined in `$PGDATA`, I'm extending the check for SSL to incorporate that as well, otherwise postgresql fails to start. 
For details see [bug 1191181](https://bugzilla.redhat.com/show_bug.cgi?id=1191181) and [postgresql docs](http://www.postgresql.org/docs/9.2/static/ssl-tcp.html).


soltysh:
@bparees PTAL.

bparees:
lgtm.

soltysh:
[merge]

PR #6078, opened 2015-02-19
Change for SSH minimum key size check

oybed:
This is the code change for issue outlined in bugzilla #1194035 - https://bugzilla.redhat.com/show_bug.cgi?id=1194035

In essence, it is adding an optional check (if configured in broker.conf) to validate the SSH key size (i.e.: number of bits used to generate the keys). The customer security scan caught this and they consider this a medium security issue. This PR will check RSA and DSA keys for size (although DSA is somewhat irrelevant as it's always going to be 1024).


oybed:
The following tests have been executed with this code change:
1. Run with MINIMUM_SSH_KEY_SIZE in broker.conf set to nothing - ensure that valid RSA and DSA keys of any size can be added
2. Run with MINIMUM_SSH_KEY_SIZE set to 1024 for both RSA and DSA - ensure that valid keys of size 1024 can be added
3. Run with MINIMUM_SSH_KEY_SIZE set to 2048 for RSA, ensure that RSA keys of size smaller than 2048 cannot be added and that an appropriate error message is displayed
4. Run with MINIMUM_SSH_KEY_SIZE set to 1025 for DSA (not valid) - ensure that a valid DSA key of size 1024 cannot be added (this test is not applicable to real world scenario, but it validates that the check for DSA does indeed work)

brenton:
Hi @oybed.  Thanks for submitting this.  @sdodson's going to help you get this committed upstream!

oybed:
Including @jwhonce

jwhonce:
@abhgupta Can you review since this impacts the Broker?  Thanks.

broker/Gemfile
```diff
@@ -11,6 +11,7 @@ gem 'systemu'
 gem 'mongoid'
 gem 'bson'
 gem 'bson_ext'
+gem 'net-ssh'
```
abhgupta: Can you confirm that this rubygem is packaged as an rpm and is being installed on the broker already? Also, we'll need a corresponding change for OpenShift Online broker configuration.

sdodson: Looks like ruby193-rubygem-net-ssh-2.7.0 is in the repos for origin and ose but it's not currently installed via any dependency so it'd need to be added to the specfile as a Requires

    Requires:      %{?scl:%scl_prefix}rubygem-net-ssh

oybed: Thanks guys - I was starting to wonder about that after I submitted the PR. I'll get the Requires line added.

controller/app/validators/key_content_validator.rb
```diff
@@ -0,0 +1,15 @@
+class KeyContentValidator < ActiveModel::Validator
+   def validate(record)
+     key = Net::SSH::KeyFactory.load_data_public_key(record.type + " " + record.content)
```
sdodson: We need some exception handling here. I'm hacking on something, but if the data passed in is not a key that Net::SSH can parse we're getting 500 errors rather than properly handling bad keys via the existing validators.

I'm thinking if the key is of a type we have a minimum size for and it fails to load the key we should return the same code we would as if it didn't meet the length requirements but with a message that the key length couldn't be validated.

@oybed were you using any particular version of net-ssh? If we can use 2.7.0 it'd be best as that's what we already have in the channels.

oybed: @sdodson Sounds good. I'll see what I can do as well. As for the net-ssh version: Yes, 2.7.0 is the one I have been using with the customer and there is no reason for changing it in the channels. However, just as an additional data point, I also had one that used net-ssh 2.9.x and it worked just as well.

oybed: @sdodson Do you happen to have an example of anything that Net:SSH can't parse? I have been trying various invalid inputs (including just changing the code to pass a random string to load_data_public_key), but in all cases the handling seems to be working OK with a proper error message on the console. I agree that we probably need some exception handling, but would like to be able to find some test cases that I can use to properly test it. Any thoughts?

sdodson: Check this gist, it's got the modified version of the ruby code that catches a lot of the exceptions, but there's two remaining.

The two offending keys are
2015-03-03 18:39:56.397 [INFO ] Parameters: {"name"=>"api", "type"=>"ssh-rsa", "content"=>"XYZ123=567[dfhhfl]"} (pid:80)

oybed: @sdodson Thank you for providing the updated code and additions to the test cases. 

One question: Looking at the exceptions handling, shouldn't the rescue line have the "NotImplementedError" as a parameter? If so, I believe it will catch everything - i.e.:
  'rescue NotImplementedError'

sdodson: We should probably add that in addition to the existing rescue as it's catching other error types already. Ruby is not my area of expertise, so I'm not too familiar with it's exception handling.

oybed: Sounds good - I'll get the PR updated with these changes.

oybed:
@sdodson Please have a look at the latest changes made to key_content_validator.rb and let me know if you're OK with this latest version. Thank you.

sdodson:
@oybed I put in a PR against your branch to add two commits, i think with those in place we should be good to go.

broker/openshift-origin-broker.spec
```diff
@@ -79,6 +79,7 @@ Requires:      %{?scl:%scl_prefix}rubygem-metaclass
 Requires:      %{?scl:%scl_prefix}rubygem-mime-types
 Requires:      %{?scl:%scl_prefix}rubygem-moped
 Requires:      %{?scl:%scl_prefix}rubygem-multi_json
+Requires:      %{?scl:%scl_prefix}rubygem-net-ssh
```
sdodson: This dependency should actually exist on the controller not the broker, so move this to controller/openshift-origin-controller.gemspec

oybed:
@sdodson - sounds great. Thanks for doing that. The require 'net/ssh' in the .rb file shouldn't be necessary once the net-ssh is fully integrated with the env, but it doesn't hurt either from what I know. Hence, it's probably best to just leave it the way you have it to ensure everything is working. 

Please let me know if there is anything else I need to do get this merged. 

@sdodson - thank you for all your help with this one.

broker/openshift-origin-broker.spec
oybed: @sdodson Thanks for making the change - appreciate it.

jwhonce:
[test]

broker/openshift-origin-broker.spec
sdodson: I'm really sorry, I meant to the controller's RPM specfile. I copied and pasted the wrong thing. This is in addition to the changes I made in the PR against your branch.

oybed: @sdodson ahh, I see. I'll update my PR with this additional change. 

Just to clarify; you want me to take it out of broker/openshift-origin-broker.spec and add it to controller/rubygem-openshift-origin-controller.spec ?

sdodson:
[test] again

PR #6079, opened 2015-02-19
Encode the response to bytes before determining the length

dashea:
The Content-Length is the length of the content in octets, while the
legnth of the string before encoding to bytes is the length in
characters. These two lengths are not necessarily the same.


danmcp:
@soltysh Review please

soltysh:
LGTM [merge]

PR #6081, opened 2015-02-20
updating links for developer resources in initial pages for cartridges

coreydaley:
Updating the links on new application pages (and a couple of other spots) to point to new resources, or point to the final destination instead of having to go through old redirects.


danmcp:
[merge]

PR #6082, opened 2015-02-23
Bug 1193193 - Fixed apc.shm_size values for php-5.4

soltysh:
@bparees for some strange reason we were dividing by 16 for php 5.4, which resulted in size/2 for `apc.shm_size`, PTAL.


bparees:
lgtm.  [test][extended:cartridge]

soltysh:
That failure is not related to my changes re-[test]-ing.

soltysh:
[merge]

PR #6091, opened 2015-03-04
RHBZ1198777: Scale inodes by 80k per GB in example configs

sdodson:


danmcp:
@sdodson Where is this request coming from?  We use these values in production.

sdodson:
@danmcp https://bugzilla.redhat.com/show_bug.cgi?id=1198777 these are just the example configuration files however, we're automatically copying those into use in online?

Here's a related bug where it looks like online switched to 80k/GB long ago.
https://bugzilla.redhat.com/show_bug.cgi?id=1031112

danmcp:
@sdodson Sorry I got mixed up.  These are not the examples I thought they were :)  And I totally missed the quota_blocks was already different.

danmcp:
[merge]

PR #6092, opened 2015-03-11
BZ1199901, BZ1199903, BZ1199904 creating and deleting monitors

calfonso:


routing-daemon/conf/routing-daemon.conf
```diff
@@ -32,7 +32,7 @@ HTTP_PORT=80
 #MONITOR_NAME=monitor_ose_%a_%n
 #MONITOR_PATH=/health_check.php
 #MONITOR_UP_CODE=1
-MONITOR_TYPE=http-ecv
+MONITOR_TYPE=http
 #MONITOR_TYPE=https-ecv
```
Miciah: Might as well delete `-ecv` here too.

routing-daemon/lib/openshift/routing/controllers/asynchronous.rb
```diff
@@ -342,13 +342,13 @@ def create_monitor monitor_name, path, up_code, type, interval, timeout
       monitors.push monitor_name
     end
 
-    def delete_monitor monitor_name, pool_name=nil
+    def delete_monitor monitor_name, pool_name=nil, type
```
Miciah: Ruby won't accept this because you cannot have a non-optional parameter after an optional parameter.  We'll need to drop `=nil` from `pool_name` to make it non-optional.

routing-daemon/lib/openshift/routing/controllers/asynchronous.rb
```diff
@@ -342,13 +342,13 @@ def create_monitor monitor_name, path, up_code, type, interval, timeout
...
       raise LBControllerException.new "Monitor not found: #{monitor_name}" unless monitors.include? monitor_name
 
       # :delete_monitor blocks
       # if the corresponding pool is being deleted (if one is specified) or
       # if the monitor is being created.
-      queue_op Operation.new(:delete_monitor, [monitor_name]), @ops.select {|op| (op.type == :create_monitor && op.operands[0] == monitor_name) || (pool_name && op.type == :delete_pool && op.operands[0] == pool_name)}
+      queue_op Operation.new(:delete_monitor, [monitor_name, type]), @ops.select {|op| (op.type == :create_monitor && op.operands[0] == monitor_name) || (pool_name && op.type == :delete_pool && op.operands[0] == pool_name)}
```
Miciah: Need to add `&& op.operands[1] == type` after `op.operands[0] == monitor_name`.

routing-daemon/lib/openshift/routing/controllers/batched.rb
```diff
@@ -104,10 +104,10 @@ def create_monitor monitor_name, path, up_code, type, interval, timeout
       @monitors.push monitor_name
     end
 
-    def delete_monitor monitor_name, pool_name=nil
+    def delete_monitor monitor_name, pool_name=nil, type
       raise LBControllerException.new "Monitor not found: #{monitor_name}" unless monitors.include? monitor_name
 
-      @lb_model.delete_monitor monitor_name if monitors.include? monitor_name
+      @lb_model.delete_monitor monitor_name, type if monitors.include? monitor_name
```
Miciah: This line shows that things are kind of tricky now.  Is it safe to assume that there will never be two monitors with the same name but different types? If not, we'll need to change `monitors` from an array of monitor names to an array of tuples of monitor name and monitor type and then modify this check accordingly.

routing-daemon/lib/openshift/routing/controllers/load_balancer.rb
```diff
@@ -101,12 +101,12 @@ def delete_pool pool_name
     def create_monitor monitor_name, path, up_code, type, interval, timeout
     end
 
-    # delete_monitor :: String, String -> undefined
+    # delete_monitor :: String, String -> undefined, String
```
Miciah: Whoops, the `, String` was inserted in the wrong place.

routing-daemon/lib/openshift/routing/controllers/asynchronous.rb
calfonso: and if pool_name is non-optional, make it && op.operands[1] == pool_name and && op.operands[2] == type

routing-daemon/lib/openshift/routing/controllers/batched.rb
calfonso: I'm not sure.

routing-daemon/lib/openshift/routing/daemon.rb
```diff
@@ -326,12 +326,13 @@ def delete_application app_name, namespace
         # having the application's name and namespace in the monitor's name).
         if @monitor_name_format && @monitor_name_format.match(/%a/) && @monitor_name_format.match(/%n/)
           monitor_name = generate_monitor_name app_name, namespace
+          monitor_path = generate_monitor_path app_name, namespace
           unless monitor_name.nil? or monitor_name.empty? or monitor_path.nil? or monitor_path.empty?
```
Miciah: Wow, that was a weird defect.  Good catch!

routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
```diff
@@ -112,10 +113,8 @@ def create_monitor monitor_name, path, up_code, type, interval, timeout
            })
```
Miciah: Don't you need to add a `.to_json` here (after the close-brace)?

routing-daemon/lib/openshift/routing/models/load_balancer.rb
```diff
@@ -42,8 +42,8 @@ def get_monitor_names
     def create_monitor monitor_name, path, up_code, type, interval, timeout
     end
 
-    # delete_monitor :: String -> undefined
-    def delete_monitor monitor_name
+    # delete_monitor :: String -> undefined, String
```
Miciah: Whoops, the `, String` was inserted in the wrong place.

Miciah:
Don't hate me, but I wonder whether it would be better not to add the `type` parameter everywhere and pass the type from `daemon.rb`, but rather to have the models read the `type` parameter from the configuration file.  Is monitor type an implementation detail of the load balancers, or is it something the daemon really should know about?

routing-daemon/lib/openshift/routing/controllers/asynchronous.rb
Miciah: I don't understand what you mean.  `op.operands` refers to the array that is the second argument to `Operation.new`.  We are checking whether there is enqueued a `:create_monitor` `Operation` with parameters such that it would cancel out the `:delete_monitor` `Operation` that we are now enqueueing.

routing-daemon/lib/openshift/routing/controllers/batched.rb
Miciah: It's probably a safe assumption given that the type is set in the configuration file and won't be changing at runtime at least.

Miciah: Might be worth adding a comment somewhere to that effect.

routing-daemon/lib/openshift/routing/controllers/asynchronous.rb
calfonso: Ah I misread it. Thanks for the clarification.

calfonso: Just to close this one out, we talked on IRC and we don't need to add the type check here in the selector since the monitory type is specified in a config file and won't be altered at runtime.

routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
calfonso: can't believe I forgot that one, thanks!!

routing-daemon/lib/openshift/routing/daemon.rb
calfonso: Thx!

calfonso:
[merge]

PR #6093, opened 2015-03-12
BZ #1200123 - Add support for external SSO

timothyh:
Initial PR for BZ#1200123


brenton:
[test]

controller/lib/openshift/sso_service.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def self.register_gear(gear)
+      notify_providers :register_gear, gear
+    end
+
+    def self.deregister_gear(gear)
+      notify_providers :deregister_gear, gear
+    end
+
+    def self.deregister_alias(app, alias_str)
```
abhgupta: Do we have no need for self.register_alias ?

controller/app/pending_ops_models/remove_alias_op.rb
```diff
@@ -31,6 +33,8 @@ def rollback
       application.save!
     end
 
+    OpenShift::SsoService.register_gear(gear) unless gear.removed
```
abhgupta: Why the inconsistency in execute vs rollback? Please add a comment why SsoService.register_alias is not a provided option. In any case, we should add comment in SsoService class to indicate that register_gear is expected to be idempotent.

controller/lib/openshift/sso_service.rb
timothyh: The self.register_gear provides equivalent functionality since it uses the gear and application objects. register_gear works for any change that adds stuff. The deregister_alias is an optimization so that the SSO back-end doesn't have to maintain state just to figure out what has changed.

controller/app/pending_ops_models/remove_alias_op.rb
timothyh: Register_gear always defines the current state of the gear/application so that it can be used for any operation. The information needed from the gear/application is the FQDN(s) that can be used to access an application, including any aliases, and the contents of some environment variables that will be used to define the associated configuration.

brenton:
Are we simply waiting for this PR to be updated with the requested comments?

abhgupta:
I haven't gone through the activemq SSO plugin in detail, but overall this PR looks good.

abhgupta:
[merge]

pmorie:
Looks like the merge problem was a flake.

[merge] again

danmcp:
[test][extended:broker]

PR #6095, opened 2015-03-12
Bug 1201485

tiwillia:
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1201485
Final selection should take the top half of the server_infos list so that nodes with the lowest consumed_capacity will be chosen

### DIFF
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -3315,7 +3315,7 @@ module OpenShift
       # Sort by district available capacity and take the best half
       server_infos = server_infos.sort_by { |server_info| (server_info.district_available_capacity ? server_info.district_available_capacity : 1) }
       # consider the top half and no less than min(4, the actual number of available)
-      server_infos = server_infos.last([4, (server_infos.length / 2).to_i].max)
+      server_infos = server_infos.first([4, (server_infos.length / 2).to_i].max)
 
       server_info = nil
       unless server_infos.empty?


plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3315,7 +3315,7 @@ def self.select_best_fit_node_impl(server_infos)
       # Sort by district available capacity and take the best half
       server_infos = server_infos.sort_by { |server_info| (server_info.district_available_capacity ? server_info.district_available_capacity : 1) }
```
sosiouxme: The first sort is by node_consumed_capacity, where "less (consumed) is better", so we take the first half.
The second sort is by district_available_capacity, where "more (available) is better", so we take the last half.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3315,7 +3315,7 @@ def self.select_best_fit_node_impl(server_infos)
...
       # consider the top half and no less than min(4, the actual number of available)
-      server_infos = server_infos.last([4, (server_infos.length / 2).to_i].max)
+      server_infos = server_infos.first([4, (server_infos.length / 2).to_i].max)
```
sosiouxme: ... so I don't think changing last to first is going to solve it properly. If all of your server_infos come from the same district, as they did in your first test case I saw, then they all have the same district_available_capacity and the sort isn't going to change anything (I would have thought it would randomize the entries, but apparently it doesn't, which is kind of nice usually, just not here), so selecting .last now gets you the "worst" nodes in terms of node capacity (from the previous sort).

But you don't want to reverse the behavior the second sort is intended for, where nodes are coming from multiple districts and you want to select districts that have capacity. Instead, I suggest augmenting it with a secondary sort to still select the nodes with best capacity in the case where they come from the same district.

danmcp:
Related: https://github.com/openshift/origin-server/pull/6099

brenton:
Online hit a different problem recently in the same code path and opt'd to solve the problem more generically.  You might want to fetch his PR and rerun your tests.  Let's close this for now.

PR #6096, opened 2015-03-13
Fix the expression to allow the nightly build

tagoh:
giving -r nightly to this script fails when $origin_release is set to nightly.
the expected expression here should be $origin_release isn't v4 nor nightly but
it was $origin_release isn't v4 or nightly. so it always fails when it isn't v4.


danmcp:
@nhr Review please.

nhr:
Looks good. Thanks, @tagoh for the PR!

danmcp:
[merge]

PR #6098, opened 2015-03-16
Use the net/http library for HTTP requests

ironcladlou:
Replace curl usage with the net/http library. Support request retries
and rubust response handling.

Fix https://bugzilla.redhat.com/show_bug.cgi?id=1196617.

### DIFF
--- a/cartridges/openshift-origin-cartridge-jenkins-client/bin/jenkins_build
+++ b/cartridges/openshift-origin-cartridge-jenkins-client/bin/jenkins_build
@@ -2,37 +2,105 @@
 
 require 'rubygems'
 require 'json'
+require 'net/https'
+require 'uri'
 
 STDOUT.sync = true
 STDERR.sync = true
 Encoding.default_external = "utf-8"
 
+# Build constants from the environment.
 @hostname = ENV['JENKINS_URL'].split("/")[-1]
+@username = ENV['JENKINS_USERNAME']
+@password = ENV['JENKINS_PASSWORD']
 @job_url = "/job/#{ENV['OPENSHIFT_APP_NAME']}-build"
 
+# A list of recoverable HTTP error codes which should indicate
+# a retry is necessary.
+RETRYABLE_ERRORS = ['502', '503', '504']
+
+# Yields until the block stops raising errors. If a retry is
+# attempted, progress and completion is reported; otherwise
+# initial success is silent.
+#
+# Sleeps for a second between retries.
+def try_forever(msg)
+  retried = false
+  loop do
+    begin
+      yield
+      puts "Done." if retried
+      return
+    rescue => e
+      unless retried
+        print "Retrying #{msg}..."
+        retried = true
+      end
+      print '.'
+      sleep 1
+    end
+  end
+end
+
+# Makes an HTTP client with SSL verification disabled.
+def make_http(uri)
+  uri = URI.parse(uri)
+  http = Net::HTTP.new(uri.host, uri.port)
+  http.use_ssl = true
+  http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+  return [http, uri]
+end
+
+# Makes a request to uri and yields the response. Will retry
+# requests whose response code is in RETRYABLE_ERRORS.
+def try_request(type, uri_string, label)
+  http, uri = make_http(uri_string)
+  case type
+  when 'GET'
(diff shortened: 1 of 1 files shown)


ironcladlou:
This PR is still a WIP.

ironcladlou:
cc @detiber @a13m @jwhonce

ironcladlou:
Should each request to Jenkins be retried forever? If not, for how long? Are there any response codes we know for a fact at a point in time can be considered immediately fatal?

ironcladlou:
cc @sferich888 @brenton @sosiouxme

detiber:
@ironcladlou I would think we could consider the 4xx errors as fatal. Probably any 5xx errors other than 502, 503, and 504 as well.

ironcladlou:
Let's turn it around: what's the inclusive list of HTTP codes which should trigger a retry? 502/503/504?

detiber:
@ironcladlou I believe that to be the case, yes. That would cover Proxy errors (gear not available), Service Unavailable (gear overwhelmed) and Gateway timeout (node temporarily unavailable)

ironcladlou:
Refactored the code further and added comprehensive error handling and reporting. I think this is in much better shape than before. I need some code review before merging.

cartridges/openshift-origin-cartridge-jenkins-client/bin/jenkins_build
```diff
@@ -2,37 +2,104 @@
...
+  http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+  return [http, uri]
+end
+
+# Makes a request to uri and yields the response. Will retry
+# requests whose response code is in RETRYABLE_ERRORS.
+def try_request(type, uri_string, label)
+  response = nil
+  try_forever(label) do
+    http, uri = make_http(uri_string)
```
jwhonce: I think make_http() should be outside the try_forever() loop or you'll be stuck in the loop should the parse fail.

ironcladlou: Good catch... I moved everything but the actual request out of the try loop.

ironcladlou:
[test]

ironcladlou:
@jwhonce give this one more review?

cartridges/openshift-origin-cartridge-jenkins-client/bin/jenkins_build
```diff
@@ -2,37 +2,105 @@
...
+  when 'POST'
+    request = Net::HTTP::Post.new(uri.request_uri)
+  end
+  request.basic_auth(@username, @password)
+
+  response = nil
+  try_forever(label) do    
+    response = http.request(request)
+    if RETRYABLE_ERRORS.include?(response.code)
+      raise "#{label}: HTTP #{response.code} #{response.message}"
```
brenton: I don't think we're printing this exception message anywhere.  It could be useful to do so.

ironcladlou: In this context, I opted to eat it during retries and print periods to indicate progress since the errors are expected.... should I dump the message amongst the progress ellipses every second?

brenton: We definitely wouldn't want this being returned to the client.  I thought maybe there was another log available.  It just looked odd to set a message that nothing ever read.

cartridges/openshift-origin-cartridge-jenkins-client/bin/jenkins_build
```diff
@@ -45,62 +113,113 @@ def is_cancelled
   queueItem = jobs_info["queueItem"]
 end
 
+# Schedules a new build.
 def schedule_build
-  status_code = `curl -s -w %{http_code} --output /dev/null -X POST --insecure https://${JENKINS_USERNAME}:${JENKINS_PASSWORD}@#{@hostname}#{@job_url}/build`
-  unless ['302', '200', '201'].include? status_code
-    puts "ERROR - Couldn't schedule job"
-    exit 292
```
jwhonce: I verified this exit code is _not_ on the broker's categorized list, so dropping it should not be an issue.

jwhonce:
LGTM :+1:

ironcladlou:
Thanks. Let's [merge] and let QE take it for a spin.

PR #6099, opened 2015-03-16
Take district less into account when filtering available servers

danmcp:


danmcp:
[test]

danmcp:
@a13m Review please

brenton:
@danmcp  Looking through the new sorting I think https://github.com/openshift/origin-server/pull/6095 should be handled by this.  Would you mind taking a quick look?

The use case is simple, customer has a single large district.  The previous sorting we were doing on district_available_capacity wasn't useful in this case and we would always prefer the same nodes.

danmcp:
@brenton Yeah this will certainly fix that case as well.

danmcp:
@sosiouxme Can you review this too please

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3307,15 +3307,26 @@ def self.select_best_fit_node_impl(server_infos)
...
-      server_infos = server_infos.sort_by { |server_info| (server_info.district_available_capacity ? server_info.district_available_capacity : 1) }
-      # consider the top half and no less than min(4, the actual number of available)
-      server_infos = server_infos.last([4, (server_infos.length / 2).to_i].max)
+      if server_infos.length > 4
+        # Take all nodes with < the average capacity.  Make sure to take at least max(4, 20% of nodes) nodes.
+        server_infos = server_infos.sort_by { |server_info| server_info.node_consumed_capacity }.reverse!
+        node_consumed_capacities = server_infos.map { |server_info| server_info.node_consumed_capacity }
+        average_consumed_capacity = node_consumed_capacities.inject(0.0) { |sum, c| sum + c } / node_consumed_capacities.length
+        min_nodes = [4, (server_infos.length * 0.2).to_i].max
+        server_infos.delete_if { |server_info| server_info.node_consumed_capacity > average_consumed_capacity && server_infos.length > min_nodes }
```
sosiouxme: The intent here is a little unclear, but delete_if doesn't actually delete anything (have to put the result back in the var) and server_infos.length wouldn't change during the loop. If the idea is to chop out everything that's above average, but not below min, then it would probably be best to sort and have the loop compare the index to the min_nodes to make sure you don't start chopping too soon.

Test case might help...

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3307,15 +3307,26 @@ def self.select_best_fit_node_impl(server_infos)
...
+
+        half_full_count = 0
+        server_infos.each do |server_info|
+          if server_info.district_available_capacity && server_info.district_available_capacity < (Rails.configuration.msg_broker[:districts][:max_capacity].to_f / 2).to_i
+            half_full_count += 1
+          end
+        end
+        half_full_ratio = half_full_count.to_f / server_infos.length
+        # Sort by district available capacity
+        server_infos = server_infos.sort_by { |server_info| (server_info.district_available_capacity ? server_info.district_available_capacity : 1) }
```
sosiouxme: why not just `server_info.district_available_capacity || 1`

danmcp: I didn't change that line but I do like yours better.  Will change.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
sosiouxme: I was completely wrong here, nevermind this.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3307,15 +3307,26 @@ def self.select_best_fit_node_impl(server_infos)
...
+        # consider the top 80% and no less than min(4, the actual number of available).  Take more than 80% if that many districts are less than 50% full.
+        server_infos = server_infos.last([4, (server_infos.length * [0.8, (1-half_full_ratio)].max).to_i].max)
```
brenton: If we had a district that was mostly empty and all the nodes were in that district I think we could hit a case where the half_full_ration is 1.  That wouldn't allow more than 80% of the nodes to be selected in this case.

danmcp: @brenton Your case is actually covered now by the reverse! above.  Even in cases when some districts are full.

brenton: Ahh, and now I see that. :)

a13m:
This looks like it will work for us.  Thanks.

danmcp:
[merge]

PR #6103, opened 2015-03-17
Don't delete archives if tar returns !0

dobbymoodge:
`tar` will return `1` if a "recoverable" error has happened, e.g. if a
file changed while it was being archived. We were erroneously deleting
the archive if this happened. Also, a "broken" archive may still be
useful, so now we only report if archiving failed, we don't delete the
archive.

This change also adds the `ARCHIVE_DESTROYED_GEARS_COMPRESSION` option,
which can be '`bzip2`', '`gzip`' or '`none`'.

Enterprise bug: 1200096
https://bugzilla.redhat.com/show_bug.cgi?id=1200096


dobbymoodge:
[test]

jwhonce:
[merge]

PR #6104, opened 2015-03-17
Take district less into account when filtering available servers

danmcp:


danmcp:
[merge]

danmcp:
@maxamillion @a13m FYI

PR #6106, opened 2015-03-19
Be more precise than +1 to avoid edge cases and use a weighted selection...

danmcp:
... favoring most available nodes

### DIFF
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -3311,9 +3311,11 @@ module OpenShift
         # Take all nodes with > the average remaining capacity.  Make sure to take at least max(4, 20% of nodes) nodes.
         server_infos.sort_by! { |server_info| server_info.node_consumed_capacity }.reverse!
         node_consumed_capacities = server_infos.map { |server_info| server_info.node_consumed_capacity }
-        average_consumed_capacity = (node_consumed_capacities.inject(0.0) { |sum, c| sum + c } / node_consumed_capacities.length) + 1
+        average_consumed_capacity = (node_consumed_capacities.inject(0.0) { |sum, c| sum + c } / node_consumed_capacities.length)
+        # Add a little so average isn't as harsh at the low end
+        cut_off_capacity = average_consumed_capacity + (1 - average_consumed_capacity/100)
         min_nodes = [4, (server_infos.length * 0.2).to_i].max
-        server_infos.delete_if { |server_info| server_info.node_consumed_capacity > average_consumed_capacity && server_infos.length > min_nodes }
+        server_infos.delete_if { |server_info| server_info.node_consumed_capacity > cut_off_capacity && server_infos.length > min_nodes }
 
         half_full_count = 0
         server_infos.each do |server_info|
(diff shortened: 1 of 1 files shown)


danmcp:
@a13m @sosiouxme @brenton Thoughts?

danmcp:
[test]

danmcp:
Thanks to @a13m for help with the algorithm.

danmcp:
[merge]

sosiouxme:
heh, i had some bikeshedding to do but was offline at the time and couldn't post it! oh well, nothing *wrong* per se...

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3311,9 +3311,11 @@ def self.select_best_fit_node_impl(server_infos)
         # Take all nodes with > the average remaining capacity.  Make sure to take at least max(4, 20% of nodes) nodes.
         server_infos.sort_by! { |server_info| server_info.node_consumed_capacity }.reverse!
         node_consumed_capacities = server_infos.map { |server_info| server_info.node_consumed_capacity }
-        average_consumed_capacity = (node_consumed_capacities.inject(0.0) { |sum, c| sum + c } / node_consumed_capacities.length) + 1
+        average_consumed_capacity = (node_consumed_capacities.inject(0.0) { |sum, c| sum + c } / node_consumed_capacities.length)
+        # Add a little so average isn't as harsh at the low end
+        cut_off_capacity = average_consumed_capacity + (1 - average_consumed_capacity/100)
```
sosiouxme: I may be missing the point here. `(1 - average_consumed_capacity/100)` will give:

* 1 when average is 0 - at which point everything is 0 so it doesn't matter which you keep or remove
* 0 or less when average is 100+ - at which point everything (or almost everything) is full so again it doesn't matter that you're removing (over-)full nodes.
* between 0 and 1 otherwise - tipping the cutoff point just above the average, which will only matter if it tips it past the next integer (I think capacity is stored as an integer). Is there really a use case where this makes a notable difference?

I don't have any particular objection other than it seems a bit more complicated/confusing than it needs to be.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -3330,8 +3332,17 @@ def self.select_best_fit_node_impl(server_infos)
...
+        server_infos.sort_by! { |server_info| server_info.node_consumed_capacity }.reverse!
+        # Weight the servers by their availability
+        # Divide by 2 gives a 3:1 ratio of most to least available selection.  Divide by 4 would give 5:1.
+        weight_skew = (server_infos.length.to_f / 2).round
+        # The most available nodes are at the end of the list and are associated with a larger portion of the sum
+        weights_sum = (((server_infos.length.to_f/2) * (server_infos.length-1)) + (weight_skew * server_infos.length)).to_i
+        random_weighted_position = rand(weights_sum)
+
+        # Map the random weighted position into its corresponding index
+        random_index = (0.5 - weight_skew + Math.sqrt((weight_skew - 0.5)**2 + (2 * random_weighted_position))).floor
```
sosiouxme: Should work fine... but given how long it took me to determine that, I would worry about support cases coming in with people having to be convinced of that, versus the simply random selection.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
danmcp: The reason the +1 was in there was to remove the rounding error at the lower end.  Say you have consumed capacity of:

5,5,5,6,6,6

Which would give an average of 5.5.  The normal algorithm would leave out 2 of the 6s.

But if you have:

99, 99, 99, 98, 98, 98

You probably do want to leave off the 99s.

And the new calculation is meant to solve both.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
danmcp: Maybe we should add more comments?

PR #6112, opened 2015-03-26
More routing-daemon and oo-admin-ctl-routing fixes related to bug 1199904

Miciah:
#### `oo-admin-ctl-routing`: Fix `delete-monitor` help text

Fix the help text for `oo-admin-ctl-routing`'s `delete-monitor` command.

The `delete-monitor` command takes arguments `<name> <type> [<pool>]`.  It makes sense for the monitor name and type to be together and for the optional argument to be last, so rather than changing ordering of arguments in the command, it makes more sense to change the help text to match the way the command actually works.
#### `oo-admin-ctl-routing`: Fix `delete-monitor` error msg

Fix the error message for `oo-admin-ctl-routing`'s `delete-monitor` command when it is given a wrong number of arguments.  The pool name is optional, but the monitor name is always required.
#### routing-daemon: Refresh monitors in case of error

Modify the controllers' `create_monitor` and `delete_monitor` methods to refresh the controller's cached list of monitors from the load balancer when the controller is told to create a monitor that already exists in the controller's list, or if it is told to delete a monitor that does not exist in its list.

After this commit, the routing daemon is better able to cope with the load-balancer's state's being modified from outside the routing daemon while the routing daemon is running.
#### routing-daemon: Try harder to create pool

Modify `OpenShift::RoutingDaemon#create_application` to rescue and log any `LBControllerException` raised in the controller's `create_monitor` method so that `create_application` will continue on to try to create the pool.

The goal of this commit is to make the routing daemon try a little harder to cope with inconsistent state.

This pull request addresses issue raised in bug 1199904.

### DIFF
--- a/routing-daemon/bin/oo-admin-ctl-routing
+++ b/routing-daemon/bin/oo-admin-ctl-routing
@@ -211,7 +211,7 @@ argvs.each do |argv|
       lb.create_monitor *argv
 
     when 'delete-monitor'
-      raise ArgumentError.new "Requires a pool name and monitor type." unless [2,3].include? argv.length
+      raise ArgumentError.new "Requires a monitor name and monitor type." unless [2,3].include? argv.length
       if 2 == argv.length
         puts "Deleting monitor #{argv[0]} of type #{argv[1]}."
         lb.delete_monitor argv[0], nil, argv[1]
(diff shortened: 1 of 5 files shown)


brenton:
[merge]

PR #6116, opened 2015-04-03
oo-admin-move eats several gears and outputs json

mmahut:
This change introduces the capacity of oo-admin-move to take a list of gears. Additional a final output in JSON can be printed out. We need both of these features for automation.

```
# oo-admin-move --gear_uuid 5217cb23dbd93c3191000152,531e6effdbd93cc514000428
[...]
Successfully moved gear with uuid '531e6effdbd93cc514000428' of app 'php' from 'ex-std-node1.stg.rhcloud.com' to 'ex-std-node2.stg.rhcloud.com'
# 
```

```
# oo-admin-move --gear_uuid 5217cb23dbd93c3191000152,531e6effdbd93cc514000428 --json | tail -1
{"5217cb23dbd93c3191000152":"Error moving gear. Old and new servers are the same: ex-std-node2.stg.rhcloud.com","531e6effdbd93cc514000428":"Error moving gear. Old and new servers are the same: ex-std-node2.stg.rhcloud.com"}
# 
```


mmahut:
[test]

broker-util/oo-admin-move
```diff
@@ -83,44 +88,49 @@ if timeout
...
+      puts ue.message
+      move_status[uuid] = ue.message
+    end
+
+    unless reply.nil?
+      puts "################# ADDITIONAL DEBUG OUTPUT #################\n#{reply.debugIO.string}\n" unless reply.debugIO.string.empty?
+      puts "################# ADDITIONAL ERROR OUTPUT #################\n#{reply.errorIO.string}\n" unless reply.errorIO.string.empty?
+    end
 end
+puts move_status.to_json if json
```
a13m: if ```--json``` is specified, it would be nice if we _only_ output json, instead of having to figure out where other output ends and json starts.  ( at least have some distinction like a quiet mode or verbose mode, so there's a way to omit the other output ).

broker-util/oo-admin-move
```diff
@@ -83,44 +88,49 @@ if timeout
...
+    puts "App UUID: #{app._id}"
+    puts "Gear UUID: #{gear._id}"
+
+    destination_container = nil
+    destination_container = OpenShift::ApplicationContainerProxy.instance(target_server_identity) if target_server_identity
+
+    reply = nil
+    begin
+      reply = gear.get_proxy.move_gear_secure(gear, destination_container, destination_district_uuid, change_district, change_region, node_profile)
+      move_status[uuid] = true
```
a13m: I would at least like to see the destination_container.id as part of the json returned here.

broker-util/oo-admin-move
```diff
@@ -83,44 +88,49 @@ if timeout
...
+    rescue OpenShift::NodeException => ne
+      puts ne.message
+      move_status[uuid] = ne.message
```
a13m: We should capture more information here; possibly the whole traceback, the exception type, and the message.

mmahut:
This was little bit painful, but the final output is like this:

```
# oo-admin-move -i ex-std-node2.stg.rhcloud.com --gear_uuid 533a89142587c83d7a0000f9,5278e5392587c84815000057 --json 
{"533a89142587c83d7a0000f9":{"login":"test@redhat.com","application":"533a89142587c83d7a0000f9","destination":{"id":"ex-std-node2.stg.rhcloud.com","district":null},"result":true},"5278e5392587c84815000057":{"login":"mmahut+stg@redhat.com","application":"5278e5392587c84815000057","destination":{"id":"ex-std-node2.stg.rhcloud.com","district":null},"result":"Error moving gear. Old and new servers are the same: ex-std-node2.stg.rhcloud.com","trace":{"code":1,"resultIO":null,"field":null,"response_code":null,"data":null}}}
#

```

a13m:
We discussed this on IRC, but just to capture it here:  it would be better to make the print_debug option an instance variable instead of passing it through as an argument to every method.

mmahut:
@mfojtik do you mind a review, please?

broker-util/oo-admin-move
```diff
@@ -83,44 +88,54 @@ if timeout
...
+    end
+
+    if app.scalable and node_profile
+      puts "Cannot specify 'node_profile' for a gear belonging to a scalable application."
+      exit 1
+    end
+
+    url = "http://#{app.name}-#{app.domain.namespace}.#{Rails.configuration.openshift[:domain_suffix]}"
+
+    puts "URL: #{url}" unless json
```
mfojtik:
```
unless json
  puts ...
  puts ...
end
```

broker-util/oo-admin-move
```diff
@@ -83,44 +88,54 @@ if timeout
...
+    destination_container = OpenShift::ApplicationContainerProxy.instance(target_server_identity) if target_server_identity
+    move_status[uuid] = { :login => app.domain.owner.login, :application => app._id, :destination => destination_container } if json
+
+    reply = nil
+    begin
+      gear.get_proxy.disable_log_debug! if json
+      reply = gear.get_proxy.move_gear_secure(gear, destination_container, destination_district_uuid, change_district, change_region, node_profile)
+      move_status[uuid].merge!({ :result => true })
+    rescue OpenShift::NodeException => ne
+      puts ne.message unless json
```
mfojtik:
```
if json
  move_status[uuid].merge....
else
  puts ne.message
end
```

mfojtik: do not overuse the 'statement condition' syntax as it makes the code hard to read.

mfojtik:
LGTM, please fix the 2 coding nits I mentioned and then I will merge this.

mfojtik:
[merge]

PR #6117, opened 2015-04-06
Bug 1207486 - stop_lock gears have status of "started"

brenton:
The previous code was:

next unless uid != '0' && command =~ /jenkins\/slave.jar/ && ppid == '1'

that's the same as:

next if !(uid != '0' && command =~ /jenkins\/slave.jar/ && ppid == '1')

or:

next if uid == '0' || command !~ /jenkins\/slave.jar/ || ppid != '1'

Which is actually not what we wanted because the jenkins slave check would
short circuit the non-deamon check in the case of stop_lock checks.  The
process would be non-root and non-jenkins related which would cuase it to be
skipped.  We only wanted to do that if the process was not a daemon:

next if uid == '0' || ppid != '1' && command !~ /jenkins\/slave.jar/

That will skip root process, skip things that aren't daemons except for the
jenkins slave process.  The version committed is simply a more commented
version of this.

### DIFF
--- a/node-util/conf/watchman/plugins.d/gear_state_plugin.rb
+++ b/node-util/conf/watchman/plugins.d/gear_state_plugin.rb
@@ -193,8 +193,12 @@ class GearStatePlugin < OpenShift::Runtime::WatchmanPlugin
       command = command.join(' ')
 
       # skip everything owned by root (for speed) and not a "daemon" (we can be fooled here)
-      # bz1134686 - but don't skip jenkins builder slaves
-      next unless uid != '0' && command =~ /jenkins\/slave.jar/ && ppid == '1'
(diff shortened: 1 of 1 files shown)


brenton:
@jwhonce Mind sanity checking this?  (We did test this in OSE)

@a13m, you might be interested in this as well.  We likely introduced a regression for https://bugzilla.redhat.com/show_bug.cgi?id=1006557 with last sprint's release.  That regression would cause stale `.stop_lock` files not to be cleaned up in many cases.

brenton:
[test]

jwhonce:
:+1:  LGTM

brenton:
[merge]

PR #6118, opened 2015-04-06
Bug 1174824 - Allow gear moves to rsync by node FQDN rather than IP

a13m:


plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -1815,7 +1815,8 @@ def move_gear_post(gear, destination_container, state_map)
       if gear.group_instance.platform.downcase == "windows"
         log_debug "DEBUG: Restoring ownership and user ACLs for Windows gear '#{gear.uuid}'"
         rsync_keyfile = Rails.configuration.auth[:rsync_keyfile]
-        log_debug `eval \`ssh-agent\`; ssh-add #{rsync_keyfile}; ssh -o StrictHostKeyChecking=no -A root@#{destination_container.get_ip_address} "/cygdrive/c/openshift/bin/oo-cmd.exe oo-admin-restore-acls --uuid:#{gear.uuid}"; exit_code=$?; ssh-agent -k;exit $exit_code`
+        destination_address = Rails.configuration.msg_broker[:node_address_by_fqdn] ? destination_container.id : destination_container.get_ip_address
```
sosiouxme: Technically, I believe id is the mcollective server_identity, which doesn't necessarily have to be a FQDN or anything at all really. That seems fine, because we have to get a name from *somewhere* and the server_identity is as likely to be meaningful (with a malleable meaning, even) as anything... it just might be clearer if the config parameter were named something else. Maybe `:node_rsync_by_id` or something.

sosiouxme:
Otherwise, in principle, looks good... I assume QE would catch any outright bugs :)

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
a13m: @sosiouxme : I realized as soon as I started testing in a devenv that calling it FQDN was probably wrong, but I didn't realize that there were cases were the ID was not a hostname at all, so thanks for raising that point.  I'll think about an appropriate name change.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -1815,7 +1815,12 @@ def move_gear_post(gear, destination_container, state_map)
       if gear.group_instance.platform.downcase == "windows"
         log_debug "DEBUG: Restoring ownership and user ACLs for Windows gear '#{gear.uuid}'"
         rsync_keyfile = Rails.configuration.auth[:rsync_keyfile]
-        log_debug `eval \`ssh-agent\`; ssh-add #{rsync_keyfile}; ssh -o StrictHostKeyChecking=no -A root@#{destination_container.get_ip_address} "/cygdrive/c/openshift/bin/oo-cmd.exe oo-admin-restore-acls --uuid:#{gear.uuid}"; exit_code=$?; ssh-agent -k;exit $exit_code`
+        destination_address = case Rails.configuration.msg_broker[:node_rsync_address]
+          when 'public_hostname' then destination_container.get_public_hostname
+          when 'server_identity' then destination_container.id
+          else destination_container.get_ip_address
+        end
```
sosiouxme: por favor, factor these out into a method call... the only thing different is the container var :)

sosiouxme:
[merge]

PR #6119, opened 2015-04-08
Added generic call script custom DNS Plugin Gem

grahamnscp:
Hi guys, created a useful generic DNS plugin that calls a script specified in the config file.  An example script that calls nsupdate is included in the conf directory too.
Creating a plugin was not well documented so also serves as an example as well as allowing a free form script to wrap any custom DNS deployment requirements.  
Hope that it would be a useful addition to the upstream project, also my first upstream pull request :)
Best regards, Graham


danmcp:
@sosiouxme Review please.

sosiouxme:
@grahamrht The concept is good, the implementation looks good at first glance, but it will have to be added to my list to review fully after a few other things are done. Sorry for the wait, thanks for getting this moving!

grahamrht:
Thanks Luke,

Best regards,
Graham

On 11/06/2015 11:44, Luke Meyer wrote:
>
> @grahamrht <https://github.com/grahamrht> The concept is good, the 
> implementation looks good at first glance, but it will have to be 
> added to my list to review fully after a few other things are done. 
> Sorry for the wait, thanks for getting this moving!
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/openshift/origin-server/pull/6119#issuecomment-111177964>.
>

-- 
Graham Hares, RHCA
Red Hat Consulting
Mobile: 07584 680158

plugins/dns/custom/conf/ose-dns-custom
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+
+# Main
+
+# Params for nsupdate
+hostname=127.0.0.1
+priv_key=/etc/openshift/example.com.key
+ttl=60
+zone=example.com
```
tiwillia: Should these parameters be read in from a configuration file? Few actual openshift environments have named installed on the same host as a broker, so this plugin won't work for most.

They should either be read in from a configuration file or passed into the script as parameters. Normally, dns plugins establish connection information in the plugin class's `initialize` method. I can see why that wasn't done here, since it would be cumbersome to pass the connection information every time the script is called. If the connection information isn't passed as parameters, they should at least be read in from a configuration file.

One issue with defining the connection parameters in this script is that this file is not marked as a configuration file in the rpm spec. If a newer version of the rpm is installed, it will overwrite the modified file with the one in the newer rpm.

plugins/dns/custom/README.md
```diff
@@ -0,0 +1,35 @@
+# Configuration
+
+This plugin provides a DNS integration for OpenShift Enterprise version 2.2.
+The plugin essentially allows calling of a local script that can be used to integrate with a remote system as required.  In the example the script calls local nsupdate commands.
+
+The configuration file for the plugin is ```/etc/openshift/plugins.d/openshift-origin-dns-custom.conf```
+
+Three variables in the configuration file define the location of the update server:
```
tiwillia: This mentions three variables in the configuration file. There looks to only be one and it doesn't define the location of the update server.

tiwillia:
@grahamrht the `openshift-origin-dns-custom-1.0.0.gem` file was included with this pr. There is no need to include the built gem, it should be removed.

Other than that and the few line notes I made, this looks good to me.

PR #6123, opened 2015-04-14
BZ1210489 -  Enable-HA registers routing appliaction DNS twice

tiwillia:
Bug 1210489
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1210489
The routing dns entry is added twice. This addition, added in d15a9d2aad33cd3ae0723ee9ac2c126c1ab1966e, is not needed as a method to create an HA application in a single API call is not implemented.

### DIFF
--- a/controller/app/models/application.rb
+++ b/controller/app/models/application.rb
@@ -1938,7 +1938,10 @@ class Application
       register_dns_op = RegisterDnsOp.new(gear_id: gear_id, prereq: [create_gear_op._id.to_s])
       register_sso_op = RegisterSsoOp.new(gear_id: gear_id, prereq: [register_dns_op._id.to_s])
       ops.push(init_gear_op, reserve_uid_op, create_gear_op, register_dns_op, register_sso_op)
-      ops.push(RegisterRoutingDnsOp.new(prereq: [register_sso_op._id.to_s])) if self.ha and Rails.configuration.openshift[:manage_ha_dns]
+      # Only register the routing dns here on application creation:
+      if self.component_instances == []
+        ops.push(RegisterRoutingDnsOp.new(prereq: [register_sso_op._id.to_s])) if self.ha and Rails.configuration.openshift[:manage_ha_dns]
+      end
 
       if additional_filesystem_gb != 0
         # FIXME move into CreateGearOp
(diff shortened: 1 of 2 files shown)


brenton:
Let's add a test for this and then merge it in.

brenton:
[test]

controller/app/models/application.rb
```diff
@@ -1938,7 +1938,10 @@ def calculate_gear_create_ops(ginst_id, gear_ids, deploy_gear_id, comp_specs, co
       register_dns_op = RegisterDnsOp.new(gear_id: gear_id, prereq: [create_gear_op._id.to_s])
       register_sso_op = RegisterSsoOp.new(gear_id: gear_id, prereq: [register_dns_op._id.to_s])
       ops.push(init_gear_op, reserve_uid_op, create_gear_op, register_dns_op, register_sso_op)
-      ops.push(RegisterRoutingDnsOp.new(prereq: [register_sso_op._id.to_s])) if self.ha and Rails.configuration.openshift[:manage_ha_dns]
+      # Only regsiter the routing dns here on app creation
```
brenton: Typo :)

brenton:
@rajatchopra, would you mind taking a quick look to see if checking the `component_instances` in this case is sufficient for determining if the other gear creation has already triggered DNS creation?

tiwillia:
@rajatchopra @brenton I'm not entirely sure checking for an empty component_instances is what we need to do here. For users, the workflow is to create the app THEN make-ha. For some of our tests, we use the 'ha' parameter on the api call to create the application as HA. The routing DNS is only registered once in the normal user workflow, but not in our tests. 

I believe this is because both gears are created at the same time, so component_instances is empty for both of them. Any ideas on a better way to ensure this op is added only once per application are appreciated.

Miciah:
@tiwillia, what happens if we simply don't push `RegisterRoutingDnsOp` in `MakeAppHaOpGroup`?

tiwillia:
@Miciah if we don't push `RegisterRoutingDnsOp` in `MakeAppHaOpGroup`, the routing DNS will never be created for an application with the check on `component_instances` in `application.rb` since the application is not normally created with ha enabled. Creating the app, then using `rhc app enable-ha` or equivalent API call is the correct workflow for users when creating an HA application. 

If we didn't do the `component_instances` check and removed `RegisterRoutingDnsOp` from `MakeAppHaOpGroup`, then `RegisterRoutingDnsOp` would be pushed for every new gear that was added to the HA application, when it should only be run once per application.

brenton:
[merge]

PR #6124, opened 2015-04-15
Attempting to fix BZ 1212020

sferich888:
Doing the same type checking that we do for the create monitor should, also allow us to resolve issues seen in miss/off configurations, when we delete monitors.


brenton:
@Miciah, here's the upstream PR.  Please have a look. :)

Miciah:
This change looks correct.  [merge]

PR #6125, opened 2015-04-16
Bug 1212614 - Various oo-admin-move issues

a13m:
- command failed when --json was not specified
- command aborted prematurely if a single gear in a list was not found
- destination server was not properly reported unless explicitly specified
- source server identity is now also captured in json output


abhgupta:
[merge] LGTM

PR #6127, opened 2015-04-17
fixed syntax error and remove unused variable 'user'

pruan-rht:
fix this bug. https://bugzilla.redhat.com/show_bug.cgi?id=1212277


pruan-rht:
this issue is fixed already by en earlier PR.

PR #6128, opened 2015-04-23
Change openshift-tc starting priority

IshentRas:
Causing issues (File : node/lib/openshift-origin-node/utils/tc.rb:105) when using bonding interfaces (bondx) as EXTERNAL_ETH_DEV as the MTU is not actually predefined before the start.


brenton:
[test]

tiwillia:
[test] failure unrelated

IshentRas:
will be reopened due to cleaning own mess

PR #6133, opened 2015-04-30
BZ1216220 - oo-snapshot fails when run outside of a gear

tiwillia:
Bug 1216220
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1216220
oo-snapshot and gear fail when run outside of a gear. Instead, these will fail gracefully with an error stating that they should not be run outside of a gear.

### DIFF
--- a/node/misc/bin/gear
+++ b/node/misc/bin/gear
@@ -61,8 +61,14 @@ def valid_git_ref?(ref)
   end
 end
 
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

brenton:
[merge]

PR #6134, opened 2015-05-05
Bug 1171815 - Add '--remove_all_env' option to remove all env variables

mfojtik:
@a13m PTAL

### DIFF
--- a/broker-util/oo-admin-ctl-domain
+++ b/broker-util/oo-admin-ctl-domain
@@ -34,6 +34,8 @@ Options:
 -v|--env_value <env var value>
    --allowed_gear_sizes <allowed gear sizes>
     A comma separated list of gear sizes allowed on the domain
+--remove_all_env
+    Confirm delete all environment variables
 -h|--help:
     Show Usage info
 USAGE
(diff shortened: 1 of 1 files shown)


mmahut:
Looks good, +1.

a13m:
"remove all env" is not an accurate description of what is happening here.  The code is removing all environment variables for a particular _component_, such as the jenkins server.  If a domain has several apps with cartridges which set domain-wide variables,  and you specify that you want to delete a jenkins variable, this will only delete all of the jenkins variables.  The variables related to other apps (the backup cartridge, for example) will remain.

Other thoughts:

* the admin should ideally be able to specify the cartridge name which "owns" the variables, rather than one of the variables. 
* we should warn the admin which variables we are going to delete, and request that they confirm

mfojtik:
@a13m what name you suggest then? "--remove_component_env" ? If I understand this properly, then the env['component_id'] is set (I dunno from where it comes...). Are you suggesting to add '--component_id' parameter as well?

a13m:
@mfojtik , the component_id comes from the cartridge instance which set the variable(s).  This is why I'm suggesting that you could have something like:

```oo-admin-ctl-domain -l <login> -n <namespace> -c env_del --cartridge jenkins```

which could look up an instance of the jenkins cartridge in a domain, get its component_id, and remove all variables containing that component_id from the domain.

brenton:
I'm not sure the term "component" is something we document for users.  It would probably be better to use terms they are familiar with and instead simply say cartridge.  While it would certainly be nice if the tool could detect known failure cases and fix them without human interaction I would be OK if the tool could simply list the environment variables on the domain and then allow the admin to pass in the ones that need to be deleted.  I think most of the time admins will know which ones need to be deleted.

A watchman plugin would likely be more suitable for catching cases where the linked `component_id` is now missing.

danmcp:
@mfojtik There isn't a normal use case for this right?  Meaning this command is only needed as a cleanup mechanism for something else that has gone wrong?

tiwillia:
After speaking with @a13m we decided this was probably best fixed through oo-admin-repair since the goal of this fix is to remove broken database entries. https://github.com/openshift/origin-server/pull/6162

PR #6139, opened 2015-05-12
oo-admin-repair should properly handle HA apps with deleted head gears

tiwillia:
Bug 1218049
Bugzilla Link https://bugzilla.redhat.com/show_bug.cgi?id=1218049
oo-admin-repair previously attempted to resolve HA applications with missing head gears as though they were simply missing a web-proxy gear. When a head gear is missing, the application cannot be recovered. oo-admin-repair should offer to cleanly remove the application rather than attempt to fix it (and break it more in the process).


brenton:
[test]

brenton:
[merge]

PR #6140, opened 2015-05-13
oo-admin-ctl-app: allow start/stop/restart of application cartridges

tiwillia:
Allow administrators to start, stop, and restart just a specific application cartridge across all gears in an application.

### DIFF
--- a/broker-util/oo-admin-ctl-app
+++ b/broker-util/oo-admin-ctl-app
@@ -117,17 +117,25 @@ def check_user_response
   end
 end
 
+
+begin
+  ci = app.component_instances.find_by(cartridge_name: cartridge)
+rescue Exception => e
+  puts "Cartridge #{cartridge} not found in the application #{app.name}"
+  exit 1
+end if cartridge
+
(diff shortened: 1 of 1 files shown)


broker-util/oo-admin-ctl-app
```diff
@@ -121,13 +121,53 @@ reply = ResultIO.new
...
+      reply.append app.stop(nil, true)
+    end
   when "restart"
-    reply.append app.restart  
+    if cartridge
+      begin
+        ci = app.component_instances.find_by(cartridge_name: cartridge)
+      rescue Exception => e
+        puts "Cartridge #{cartridge} not found in the application #{app.name}"
+        exit 1
```
brenton: See if you can find a reasonable way to avoid all the duplicate code.  You could simply move the begin/rescue bits to a function or you could get fancier with abstract all of it so that all you need to do is tell the function to start/stop/restart and it would do everything here.

tiwillia:
[test]

brenton:
[merge]

PR #6141, opened 2015-05-15
Bug 1221786 - Provide a way for users to set a usage_account_id

brenton:


brenton:
[test]

brenton:
@liggitt, I think @sferich888 mentioned this to you yesterday.  Is this all that's involved?

broker-util/oo-admin-ctl-user
```diff
@@ -616,6 +641,13 @@ if args["--setplanquantity"]
   plan_quantity = args["--setplanquantity"].to_i
 end
 
+if args["--setusageaccountid"]
+  unless args["--setusageaccountid"] =~ /^[0-9]+$/
+    $stderr.puts "ERROR: Usage account ID must be a positive integer"
+    exit 1
+  end
+  usage_account_id = args["--setusageaccountid"].to_i
```
liggitt: the type is actually a string in mongo, but looks like some code makes assumptions that it's an integer... for example, admin_helper.rb:
```
$user_hash.each { |k,v| billing_user_hash[k] = v if v['usage_account_id'] and (v['usage_account_id'].to_i > 0) }
```
requiring an int here is probably a good first pass

liggitt:
@brenton, looks good as far as I know

brenton:
[merge]

PR #6145, opened 2015-05-18
Install cpan dependencies in listed order

Miciah:
Respect the order in which dependencies are listed in `.openshift/cpan.txt` or `deplist.txt`, and install them prior to detected dependencies.

Uniqueness is still enforced by excluding subsequent duplicates.

An instance where ordering matters is with Dancer and Plack::Builder.  Even though installing Dancer causes Plack::Builder to be installed, the installation of Dancer fails unless Plack::Builder is already installed. Thus it is required either to retry the installation of Dancer or to explicitly install Plack::Builder first (which can be done by listing it before Dancer in `.openshift/cpan.txt`).

### DIFF
--- a/cartridges/openshift-origin-cartridge-perl/bin/build
+++ b/cartridges/openshift-origin-cartridge-perl/bin/build
@@ -43,7 +43,7 @@ then
 
   DEPS=$(find $OPENSHIFT_REPO_DIR -type f \( -name \*\.pm -o -name \*\.pl \); )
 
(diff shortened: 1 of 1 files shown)


Miciah:
[test][extended:cartridge]

Miciah:
[test] again because all the test failures appear to involve JBoss and not anything in this PR.

Miciah:
https://github.com/openshift/origin-server/pull/6151 supersedes this PR (if I had been thinking more clearly, I would have simply updated this PR—whoops!).

PR #6146, opened 2015-05-18
Change the priority number for openshift-tc

IshentRas:
- Change the priority number to be higher than the network interfaces ( number 10).
- Causing issues (File : node/lib/openshift-origin-node/utils/tc.rb:105) when using bonding interfaces (bondx) as EXTERNAL_ETH_DEV as the MTU is not actually predefined before the start.

### DIFF
--- a/node/misc/init/openshift-tc
+++ b/node/misc/init/openshift-tc
@@ -6,7 +6,7 @@
 # Liberally excerpted from an example by Scott Seong
 #   http://www.topwebhosts.org/tools/traffic-control.php
 #
-# chkconfig:    345 7 93
+# chkconfig:    345 11 89
 #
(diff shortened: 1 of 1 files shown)


brenton:
[test]

tiwillia:
We should probably also have openshift-tc stop before network stops. Network has a stop priority of 90, so a stop priority of 89 on openshift-tc should be perfect.

IshentRas:
I agree with you +1, I am going to modify this.

IshentRas:
@tiwillia :+1:

tiwillia:
[test]

tiwillia:
@brenton this is ready to merge in my opinion.

brenton:
[merge]

PR #6149, opened 2015-05-20
Specify addition rubygems in the openshift-console, rathen than the a…

tiwillia:
…dmin-console

Bug 1094454
A change made for 1094454 (pr 5920) allowed additional rubygems for the openshift management console, rather than the openshift console.


brenton:
[merge]

PR #6150, opened 2015-05-20
Read configuration files from Gemfile as a binary

tiwillia:
Bug 1094454
https://bugzilla.redhat.com/show_bug.cgi?id=1094454
External data must be read in as binary to avoid encoding conflicts in some environments where the encoding of the configuration files does not match the default encoding for ruby.

The problem became evident on an environment with a UTF-8 encoded broker.conf file:

  # file /etc/openshift/broker.conf
  /etc/openshift/broker.conf: UTF-8 Unicode English text

### DIFF
--- a/broker/Gemfile
+++ b/broker/Gemfile
@@ -63,18 +63,26 @@ else
 end
 
 # Load extra gems specified in the broker configuration file
-conf_file_path = (ENV['OPENSHIFT_CONF_DIR'] || "/etc/openshift/") + 'broker.conf'
-if File.exists?(conf_file_path)
-  conf_file = File.open(conf_file_path)
-  additional_gems = conf_file.read[/^ADDITIONAL_RUBYGEMS\s*=\s*(.+)/, 1]
-  if additional_gems
-    gem_list = additional_gems.split(" ").uniq
-    gem_list.each do |name|
-      gem name.gsub(/["']/, "")
+begin
+  conf_file_path = (ENV['OPENSHIFT_CONF_DIR'] || "/etc/openshift/") + 'broker.conf'
+  if File.exists?(conf_file_path)
+    # Read file as binary to avoid encoding issues when configuration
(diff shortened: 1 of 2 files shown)


broker/Gemfile
```diff
@@ -65,7 +65,7 @@ end
 # Load extra gems specified in the broker configuration file
 conf_file_path = (ENV['OPENSHIFT_CONF_DIR'] || "/etc/openshift/") + 'broker.conf'
 if File.exists?(conf_file_path)
-  conf_file = File.open(conf_file_path)
+  conf_file = File.open(conf_file_path, 'rb')
```
brenton: I tested this out and it seems to work OK on the weird EC2 node where we saw the problem.  I would _highly_ recommend putting a comment here.  This is still a bit of a mystery.  On my local dev system:

~~~
# file /etc/openshift/broker.conf
/etc/openshift/broker.conf: ASCII English text
~~~

On the system where it's "broken":

~~~
# file /etc/openshift/broker.conf
/etc/openshift/broker.conf: UTF-8 Unicode English text
~~~

broker/Gemfile
```diff
@@ -65,7 +65,7 @@ end
 # Load extra gems specified in the broker configuration file
```
brenton: Given the weirdness we saw here I would wrap this whole thing in a begin/rescue block.  We can't afford some edge case breaking things for customers who don't need this feature.

brenton: Also, in the rescue block print the error to $stderr.  Also change the `puts` in line 77 to `$stderr.puts`.  That will print out the error on `service openshift-broker restart`.  I have no idea where stdout is going.

tiwillia:
@brenton, I made the changes as per your recommendations. Tested with UTF-8 encoding on the configuration files and the error messages show up. Could you give this one more look?

brenton:
works on my system. [merge]

PR #6154, opened 2015-05-28
Bugs 1217572 and 1226061: Add F5 device-group sync, fix ssh/scp error handling

Miciah:
##### routing-daemon: F5: Use configured SSH user

This commit fixes bug 1227501.
##### routing-daemon: F5: check for and log SSH errors

`F5IControlRestLoadBalancerModel`: Add `run` method that runs the given command, captures stdout and stderr, and raises an exception if the command fails.

`F5IControlRestLoadBalancerModel#add_ssl`: Use the new `run` method to check for, log, and handle errors from the `ssh` and `scp` commands.  Modify the logging output to be neater and more consistent.

This commit fixes bug 1226061.
##### routing-daemon: F5: Fix variable names & comments

`F5IControlRestLoadBalancerModel#add_ssl`: Change `certfname` and `keyfname` to `certfile` and `keyfile` so as better to reflect the variables' contents, and fix some grammatical errors in a comment.
##### routing-daemon: Delete `read_config` in controllers

`SimpleLoadBalancerController`, `BatchedLoadBalancerController`, `AsyncLoadBalancerController`: Delete `@virtual_server_name` initialization in `read_config` because we never use `@virtual_server_name`, and delete `read_config` because it doesn't initialize anything else.
##### routing-daemon: controllers: invoke model update

`AsyncLoadBalancerController#update`, `BatchedLoadBalancerController#update`: Invoke the model's `update` method if it exists.

Note that `SimpleLoadBalancerController` does not define its own `update` method, so it inherits the `update` method from `LoadBalancerController`, which also invokes the model's `update` method.

Note too that the only model that has an `update` method is `NginxLoadBalancerModel`, and we use `SimpleLoadBalancerController` with this model, so we haven't been missing the update method in the other controllers.
##### routing-daemon: F5: Sync device-group on update

`routing-daemon.conf`: Add commented-out example `BIGIP_DEVICE_GROUP` setting.

`F5IControlRestLoadBalancerModel#read_config`: Read `BIGIP_DEVICE_GROUP` setting and assign it to `@device_group`.

`F5IControlRestLoadBalancerModel`: Add `update` method that synchronizes the device group if `@device_group` is set.

This commit fixes bug 1217572.

PR #6156, opened 2015-06-02
Fixing auth user issue.

rjhowe:
If using this this config in etc/openshift/routing-daemon.conf 

BIGIP_HOST=F5LTM
BIGIP_USERNAME=ose
BIGIP_PASSWORD=xxxxxxx
BIGIP_SSHKEY=/etc/openshift/bigip.key

After setting up and installing a custom alias with a certificate and key, the routing daemon reports that it’s removing both temp key and cert:

D, [2015-05-27T09:32:04.687843 #25771] DEBUG -- : LTM removing temporary alias certificate. rm -f /var/tmp/www.example.com.crt
D, [2015-05-27T09:32:09.921415 #25771] DEBUG -- : LTM removing temporary alias key. rm -f /var/tmp/www.example.com.key

However, the removal of the temp key does not work. This is verified by looking in /var/tmp/ on the BigIP. 

This is due to "admin" being hard coded to the commands and using the BIGIP_SSHKEY value. This will fail since the ssh key is most likely configure to the username defined in the config file.  

Changed it to be "username" which is defined as BIGIP_USERNAME using BIGIP_SSHKEY. This will work, User must still have terminal settings in F5 = advanced shell"

### DIFF
--- a/routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
+++ b/routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
@@ -247,8 +247,8 @@ module OpenShift
 
         # scp cert and to F5 LTM (requires ssh key to be in authorized_keys on the F5 LTM
         @logger.debug("Copying certificate and key for alias #{alias_str} for pool #{pool_name} to LTM host")
-        result = `scp -o StrictHostKeyChecking=no -o PasswordAuthentication=no -o VerifyHostKeyDNS=no -o UserKnownHostsFile=/dev/null -i #{@ssh_private_key} #{certfname.path} admin@#{@host}:/var/tmp/#{alias_str}.crt`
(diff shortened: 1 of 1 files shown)


tiwillia:
LGTM. @Miciah could you give this a look? It appears to be related to your pr https://github.com/openshift/origin-server/pull/6154

Miciah:
Ooh, good catch.  Do we have a Bugzilla report for that?

@rjhowe, this PR will have merge conflicts with the PR that @tiwillia referenced, so I'll rebase your commit and incorporate it into that PR, if that's OK.

tiwillia:
Bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1227501

tiwillia:
@danmcp we can close this, since @Miciah will be incorporating this into the F5 changes.

PR #6162, opened 2015-06-11
oo-admin-repair: Fix orphaned domain environment variables

tiwillia:
Bug 1171815
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1171815
It is possible, in some cases, for a domain environment variable to remain part of the domain after the related component has been removed. oo-admin-repair should be able to detect and clear these orphaned environment variables

### DIFF
--- a/broker-util/oo-admin-repair
+++ b/broker-util/oo-admin-repair
@@ -34,6 +34,7 @@ The following issues can be fixed with this script:
   - apps due to server node down/decommissioned
   - domains containing invalid gear sizes in their allowed gear sizes list
   - users containing invalid gear sizes in their capabilities
+  - existing domain environment variables for components that no longer exist
 
 == Usage
 
@@ -54,6 +55,10 @@ Options:
     Fix or cleanup apps due to server node down/decommissioned
 --gear-sizes
     Fix gear sizes in domains users and to remove invalid gear sizes
+--orphaned-envs
+    Fix domains that have environment variables for missing components
+--domain
+    Specify a specific domain to fix (only used with --orphaned-envs)
 --usage
     Fix resource usage mismatches in mongo
 -h|--help
@@ -72,6 +77,8 @@ begin
     ["--consumed-gears",         GetoptLong::NO_ARGUMENT],
     ["--removed-nodes",          GetoptLong::NO_ARGUMENT],
     ["--gear-sizes",             GetoptLong::NO_ARGUMENT],
+    ["--orphaned-envs",          GetoptLong::NO_ARGUMENT],
+    ["--domain",                 GetoptLong::REQUIRED_ARGUMENT], # Only used with --orphaned-envs
     ["--confirm",                GetoptLong::REQUIRED_ARGUMENT], # Only used for test automation
     ["--usage",                  GetoptLong::NO_ARGUMENT],
     ["--help",             "-h", GetoptLong::NO_ARGUMENT]
@@ -90,9 +97,11 @@ fix_consumed_gears = args["--consumed-gears"]
 fix_removed_nodes = args["--removed-nodes"]
 fix_usage = args["--usage"]
 fix_gear_sizes = args["--gear-sizes"]
+fix_orphaned_envs = args["--orphaned-envs"]
 
 if fix_ssh_keys.nil? and fix_district_uids.nil? and fix_consumed_gears.nil? and
-   fix_removed_nodes.nil? and fix_usage.nil? and fix_gear_sizes.nil?
+   fix_removed_nodes.nil? and fix_usage.nil? and fix_gear_sizes.nil? and
+   fix_orphaned_envs.nil?
   puts "You must specify at least one item to fix."
   usage
 end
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

broker-util/oo-admin-repair
```diff
@@ -403,6 +418,37 @@ def find_unresponsive_servers(available_servers, auto_confirm)
...
+      print_message "Unable to find domain. The domain #{namespace} does not exist. Skipping..."
+    rescue Exception => e
+      print_message "Unable to find domain #{namespace} with error: #{e.message}."
+    end
+    # Find all component_ids for every application in this domain
+    component_ids = domains[0].applications.map {|a| a.component_instances.map {|ci| ci._id}}.flatten
+  else
+    domains = Domain.all
+    # find all component_ids for every application
+    component_ids = Application.all.map {|a| a.component_instances.map {|ci| ci._id}}.flatten
```
brenton: I worry a little bit about an array that will contain all the applications in an environment as big as OpenShift Online.  It's seems OK to run this to repair a given Domain but I worry about the admin that runs this thinking it will fix up everything only to see if OOM or impact the performance on the machine running the tool.

There's probably a much more efficient query you could do in Mongo to get just the data you want.  A simplistic approach would be `Application.only('component_instances._id').all`.  The weird thing is that you will get back a Application objects where just about everything is nil except for the component_instances _id.  That will probably save quite a bit of system resources.  I would play around with it a bit to see if you could find something even better.

brenton: Alternatively, unless @a13m thinks they will run this to search the entirety of OpenShift Online for orphaned variables instead of simply repairing domains as they are reported you could just run a quick query to see how many domains or applications exist.  If it's determined to be "too big" then you could just print a warning telling the admin to run it against explicit domains.  That would be the lame way to approach this problem but it might the real world use cases right now.

tiwillia: I'm pretty sure I can reduce the load on the database as much as possible by using `Application.only(:component_instances).pluck(:_id)`, which should return *only* the array we need from the database. I'm not sure if we will still need a warning when the query is going to be "too large" to take a reasonable amount of time.

@a13m could you provide any input on this? Would the previously mentioned query be too large of a query to perform within a reasonable time on very large environments like Online? If so, do you have a guesstimate of how many applications would be considered "too large" of a query?

a13m: Constructing objects for all domains or apps is definitely not ideal in OpenShift Online.  We _do_ want to be able to validate all domains, rather than waiting for people to report a problem, though.   For one data point, I ran this:

db.domains.find({ 'env_vars.component_id': { $exists: true }  }).count()

without divulging numbers, I can say that this result represents 0.3% of our domains.  That's still a large number, but perhaps manageable if you only pull the ids.  I kicked around mongo commands for doing this.  I'm pretty unskilled at mongo, but I managed this:

```
cids = db.domains.aggregate({ $project: { cid: '$env_vars.component_id' } }, { $unwind: "$cid" }, { $group: { _id: 'cid', res: {$addToSet:'$cid'} } }); true
cids2 = db.applications.aggregate({ $match: { 'component_instances._id': { $in: res } } }, { $project: { cid: '$component_instances._id' } }, { $unwind: "$cid" }, { $group: { _id: 'cid', res: {$addToSet:'$cid'} } }); true
valid_ids = []
cids2.result[0].res.forEach(function (key) {  valid_ids.push(key.str); })
orphan_ids = []
res.forEach(function (key) { if (valid_ids.indexOf(key.str) == -1) { orphan_ids.push(key); } })
```

The two 'aggregate' commands are the important mongo part.  The remaining lines are really just comparing object ids in two lists, and not very efficiently.  it would be better done using set difference in ruby.

But just to validate the correctness of the list:
```
libra:PRIMARY> orphan_ids.length
121
libra:PRIMARY> db.applications.find({ 'component_instances._id': { $in: orphan_ids } })
libra:PRIMARY> db.domains.find({ 'env_vars.component_id': { $in: orphan_ids } }).count()
120
```

so that all makes sense.  So I think this is the right approach:

1) find domains by those that have env vars with a component id, and pull out the component id.
2) find only applications matching the component ids referenced
3) subtract the ids found in step 2 from the ids found in step 1
4) what's left are your orphans.  Iterate over the domains containing those component ids and fix them.

tiwillia:
@a13m Thank you very much for the detailed review! I've made changes similar to the aggregate functions you mentioned. Would you be able to review and test these in online? I can provide a small test script to time how long the queries will take if needed.

tiwillia:
Thanks to @a13m, we tested the db queires in online with excellent results:

user | system | total | real
-------|----------|-------|------
0.310000 | 0.020000 | 0.330000 | (  3.738470)
0.200000 | 0.040000 | 0.240000 | (  4.965202)

About 8 seconds to run both queries. This should be ready for merge.

brenton:
Great work on this. [merge]

PR #6163, opened 2015-06-11
Use nsupdate since oo-register-dns is not available yet

tiwillia:
Instruct a user to use nsupdate, as the previous paragraph states, instead of oo-register-dns. oo-register-dns (provided by the openshift-origin-broker-util rpm) is not yet available at this stage in the documentation.


nhr:
Reviewed; [merge]

PR #6164, opened 2015-06-11
Resolve race condition for cartridges activated within the same second

tiwillia:
Bug 1191283
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1191283

Using `:priority.lt => self.priority` in the `where` method would not find types with a priority set within the same second of this types priority. Mongo compares priorities at the 'second' granular level while ruby appears to compare DateTime objects at the millisecond level. 

If two cartridges had a priority set within the same second, mongo would consider them equivalent. We are now doing the comparison in Ruby, where the priority (being of type DateTime) is compared at the millisecond level.

### DIFF
--- a/controller/app/models/cartridge_type.rb
+++ b/controller/app/models/cartridge_type.rb
@@ -111,7 +111,10 @@ class CartridgeType
   def activate
     self.priority = Time.now
     if self.save
-      self.class.where(name: name, :priority.lt => self.priority).set(:priority, nil)
+      self.reload
+      # find all other types of the same name activated before this type and set their priority to nil
+      old_types = self.class.where(:name => name).select {|type| type.priority and type.priority < self.priority }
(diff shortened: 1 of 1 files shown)


controller/app/models/cartridge_type.rb
```diff
@@ -111,7 +111,10 @@ def self.cartridge_attributes(source, url=nil, text=nil)
   def activate
     self.priority = Time.now
     if self.save
-      self.class.where(name: name, :priority.lt => self.priority).set(:priority, nil)
+      self.reload
```
tiwillia: Without reloading the object, `self.priority` would not be equal to `self.class.find(self.id).priority`. I'm not sure if this is the best way to resolve that problem and would appreciate any input.

tiwillia:
[test]

abhgupta:
[merge]

PR #6166, opened 2015-06-12
Use gear uuid instead of login name to describe arg

danmcp:
Improves https://bugzilla.redhat.com/show_bug.cgi?id=1177100


danmcp:
[merge]

PR #6170, opened 2015-06-16
Bug 1230911 Validate gear frontend configurations for vhost plugin

tiwillia:
Bug 1230911
https://bugzilla.redhat.com/show_bug.cgi?id=1230911

Validate that gear frontend configurations exist when using the apache-vhost frontend plugin.

### DIFF
--- a/node-util/sbin/oo-accept-node
+++ b/node-util/sbin/oo-accept-node
@@ -634,6 +634,7 @@ def check_system_httpd_configs
   return unless http_plugins.include?('openshift-origin-frontend-nodejs-websocket')
 
   mod_rewrite = http_plugins.include? 'openshift-origin-frontend-apache-mod-rewrite'
+  mod_vhost = http_plugins.include? 'openshift-origin-frontend-apache-vhost'
   ['aliases.db', 'idler.db', 'nodes.db', 'sts.db'].each do |db|
     begin
       File.open(File.join($CONF.get('OPENSHIFT_HTTP_CONF_DIR'), db))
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

tiwillia:
[test] again, although I may have only resolved one of the two errors

tiwillia:
The tests are failing because an issue that appears to have been present for some time is now being picked up by oo-accept-node. A gear is created with an embedded cartridge (specifically mysql-5.1) as the primary cartridge. This gear has no vhost configurations, so oo-accept-node is complaining at the end of the tests.

Through investigation, I can tell that the gear is created during the *"setting storage with scaled cartridge and embedded database"* console (or site) test. The issue is only seen when an app is created during a test through `with_storage_app(:scale => true, :db => true)`. Specifically, the improper gear is created when `add_database()` is called on the app.

Still investigating. There will likely need to be a fix created for this other issue before tests will pass again after this commit to `oo-accept-node`.

tiwillia:
[test]

tiwillia:
@jwhonce Can you review please?

jwhonce:
@tiwillia LGTM if you add ```Bug 1230911``` as a prefix to PR title Rosie will update the associated bug when this PR is merged.

tiwillia:
@jwhonce Thanks for the review and the tip! I didn't realize that is how the bot picked up the association (I always though it was the bug reference in the pr comment).

As I don't have merge privileges, I'll still need someone to `[merge]` for me.

jwhonce:
[merge]

PR #6172, opened 2015-06-17
Ensure proper quota is imposed when moving gears across node profiles

tiwillia:
Bug 1229300
https://bugzilla.redhat.com/show_bug.cgi?id=1229300
When moving gears across node profiles, ensure that the quota for the new node profile is imposed on the gear during the move.

### DIFF
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -1992,19 +1992,49 @@ module OpenShift
         raise OpenShift::UserException.new("Error moving gear. Move gear only allowed from non-districted/districted node to districted node.")
       end
 
+      # get the quota usage and limits from the existing gear
+      gear_quota = get_quota(gear)
+
       destination_node_profile = destination_container.get_node_profile
-      if source_container.get_node_profile != destination_node_profile and app.scalable
-        log_debug "Cannot change node_profile for *scalable* application gear - this operation is not supported. The destination container's node profile is #{destination_node_profile}, while the gear's node_profile is #{gear.group_instance.gear_size}"
-        raise OpenShift::UserException.new("Error moving gear. Cannot change node profile for *scalable* app.", 1)
+      if source_container.get_node_profile != destination_node_profile
+        if app.scalable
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

brenton:
This looks good to me.  I could find any oo-admin-move related tests (which sort of makes sense given that the tests run on a single system).  I guess we'd just have to have QE test this as part of their suite.

a13m:
@tiwillia , this patch does not appear to consider the case where a user has elected to add extra storage to the gear, as they are allowed to do in OpenShift Online.

tiwillia:
@a13m , thanks for pointing that out. I've made changes so that any additional storage is taken into consideration. Could you review one more time?

tiwillia:
@abhgupta I got the 'OK' from @a13m several days ago. Could you review/merge?

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -1992,20 +1992,50 @@ def move_gear(gear, destination_container, destination_district_uuid, change_dis
...
+
+          # set destination container quotas
+          quota_blocks = Integer(destination_quota[:quota_blocks])
+          quota_files = Integer(destination_quota[:quota_files])
+
+          # determine additional storage, if any
+          # subtract source container quota from source node quota
+          # add the difference to destination quota
+          additional_blocks = Integer(gear_quota[3]) - get_quota_blocks
+          additional_files = Integer(gear_quota[6]) - get_quota_files
```
abhgupta: I'd be more comfortable if we used source_container.get_quota_blocks and source_container.get_quota_files in the lines above. While the source container is implied to be the current container, it is best not to make this assumption, especially since the source_container is available and explicitly specified in line 1999 above.

abhgupta:
@tiwillia LGTM, other than the one comment I made

tiwillia:
@abhgupta thanks for the excellent suggestion. I've made the recommended change. Could you review again and merge if it is acceptable?

abhgupta:
[merge]

tiwillia:
@abhgupta can you re-merge? Looks like an example of the merge queue failures.

tiwillia:
re-[merge]

PR #6173, opened 2015-06-17
Bug 1232907 - Only unidle on start by user

jwhonce:
- See also: https://bugzilla.redhat.com/show_bug.cgi?id=966790


jwhonce:
@a13m @ironcladlou Please review.  I'm not sure why we didn't have user_initiated set before...

ironcladlou:
LGTM, seems like this script (an admin/systems script) should honor locks via user_initiated=false.

jwhonce:
[merge]

jwhonce:
[merge]again

PR #6179, opened 2015-06-25
Incorrect self-documents link in README.md for markers and cron under .openshift

bparees:
https://bugzilla.redhat.com/show_bug.cgi?id=1111524


bparees:
[merge]

PR #6180, opened 2015-06-25
Missing the default value of ServerLimit and MaxClients in httpd.conf for php cartridge

bparees:
https://bugzilla.redhat.com/show_bug.cgi?id=1153616


bparees:
[merge]

PR #6184, opened 2015-06-30
BZ1225291: Additional fix

jhadvig:
Since different versions of python use different version of pip, the --trusted-host option is only available in python-3.3 for which the BZ1225291 was reported.
@soltysh PTAL


soltysh:
LGTM [merge]

PR #6197, opened 2015-07-14
Use uft8mb4 to support 4-byte unicode characters

### LINKED ISSUE #6169: Character set should be `utf8mb4`
I notice when I use the Openshift MySQL cartridge, that the character set is set to `utf8`. This is understandable as MySQL really made a mess here, but it really should be `utf8mb4` as the `utf8` that MySQL talks about is really a *MySQL utf8*...

*Real* utf8 has [max. 4 bytes per character](https://stijndewitt.wordpress.com/2014/08/09/max-bytes-in-a-utf-8-char/)
*MySQL utf8' has only 3 bytes per character and [can't encode all of Unicode](https://mathiasbynens.be/notes/mysql-utf8mb4)

If you want *full* Unicode support (and who doesn't) you should set the character set to [utf8mb4](https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf8mb4.html) in MySQL.

tiwillia:
Bug 1186681
https://bugzilla.redhat.com/show_bug.cgi?id=1186681
The utf8 character encoding in mysql supports unicode characters with 1-3 bytes. 4-byte unicode characters such as emojis require utf8mb4 encoding in mysql.

Fixes #6169 

Since the encoding is set for a database in `post_install`, this update shouldn't interfere with existing mysql cartridges. 

This pr also adds several environment variables. These are to set a default character set for client and server as well as to set a collation server and whether a client can specify a character set to use with its connection.

### DIFF
--- a/cartridges/openshift-origin-cartridge-mysql/bin/post_install
+++ b/cartridges/openshift-origin-cartridge-mysql/bin/post_install
@@ -11,7 +11,7 @@ password=$OPENSHIFT_MYSQL_DB_PASSWORD
 # TODO: Should be we running mysql_secure_installation
 
 echo "drop database test;
-create database \`${dbname}\` DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_general_ci;" | mysql_context "/usr/bin/mysql -u root -S '$socket_file'" > /dev/null || error 'Failed to create database' 188
+create database \`${dbname}\` DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci;" | mysql_context "/usr/bin/mysql -u root -S '$socket_file'" > /dev/null || error 'Failed to create database' 188
 
 echo "
 delete from user;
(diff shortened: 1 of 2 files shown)


tiwillia:
@mfojtik could you review? (you are listed as the mysql sme, lmk if I should ping someone else)

I'm unsure about the defaults in the `my.cnf.erb` file and whether these will impact existing mysql cartridges.

mfojtik:
@tiwillia can you try to create app with current config (or tweak the setting back to current default), create some sample data and then switch it to utf8mb4?

I guess I will want to see prove that this change won't explode data in existing databases :-)

tiwillia:
Since the database is created in `post_install`, the only change an existing application will see is additional configuration directives in `~/mysql/conf/my.cnf`. I can verify that these changes don't explode or inhibit an existing database. I can still use the database, access it through `mysql`, and do a `mysqldump`.

However, I found an issue creating a rails application using the `utf8mb4` type that might be an issue. Reviewing now, going to go ahead and [test][extended:cartridge]

PR #6198, opened 2015-07-14
Require only SCL packages for centos and rhel installs

### LINKED ISSUE #5759: unable to install MariaDB cartridge on CentOS v6.5 (spec file needs to be updated)
If you run OO m4 on Centos v6.5 and try to install MariaDB cartridge, it will fail (missing dependencies, mariadb-server, mariadb-devel)

the problem should be fixed in openshift-origin-cartridge-mariadb.spec (mariadb-server is always required, should be required only on Fedora & RHEL7)

```
Requires:      mariadb-server
Requires:      mariadb-devel

# For RHEL6 install mysql55 from SCL
%if 0%{?rhel}
Requires:      mariadb55
Requires:      mariadb55-mariadb-devel
%endif
````

tiwillia:
The mariadb-server and related packages are only available through SCL for CentOS 6 and RHEL 6 installations. They should only be required for fedora installs. Additionally, the SCL packages are not required for fedora installs.

Fixes #5759

### DIFF
--- a/cartridges/openshift-origin-cartridge-mariadb/openshift-origin-cartridge-mariadb.spec
+++ b/cartridges/openshift-origin-cartridge-mariadb/openshift-origin-cartridge-mariadb.spec
@@ -8,13 +8,13 @@ Group:         Network/Daemons
 License:       ASL 2.0
 URL:           http://www.openshift.com
 Source0:       http://mirror.openshift.com/pub/openshift-origin/source/%{name}/%{name}-%{version}.tar.gz
-Requires:      mariadb-server
-Requires:      mariadb-devel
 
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

abhgupta:
@tiwillia Is this waiting on some review?

tiwillia:
Yup, looks like I lost tack of this one.

@mfojtik could you review please?

mfojtik:
@tiwillia I would prefer this to be reviewed by somebody from operation team like @mmahut

mmahut:
This looks fine, +1.

mfojtik:
[merge]

PR #6200, opened 2015-07-20
Bug 1214087 - return non-zero on all exceptions in oo-admin-move

a13m:
These exit codes were removed when oo-admin-move was changed to take multiple gears as input.


tiwillia:
LGTM, @danmcp can we merge this?

danmcp:
[merge]

PR #6201, opened 2015-07-20
Bug 1241660 Ensure ignored servers does not include all servers only after filtering

tiwillia:
https://bugzilla.redhat.com/show_bug.cgi?id=1241660

Compare server_infos to least_preferred_servers only after server_infos has been filtered. Previously, the comparison was made before filtering, so it was possible for server_infos to contain only servers listed in least_preferred_servers, resulting in server_infos being returned as an empty array.

Also removed logic to only compare least_preferred_zone_ids if least_preferred_servers was not to be ignored, since it is possible for server_infos to contain all of the least_preferred_servers and more at that stage.

Finally, added a bit of logic to ensure that at least one server still exists in server_infos when removing the least preferred servers.


tiwillia:
@danmcp, this should resolve the edge case we discussed. Could you review?

danmcp:
[merge]

PR #6205, opened 2015-07-21
Ensure gears are deleted after simulating node failure

tiwillia:
Ensure gears created during the simulation of a node failure and repair are deleted from the node. Gears are created, then never deleted when oo-admin-repair repairs a missing head gear for an HA application. These cause oo-accept-node to throw errors and the extended broker test suites to fail.


tiwillia:
[test][extended:broker]

dobbymoodge:
lgtm +1

Miciah:
[merge]

PR #6212, opened 2015-08-05
Bug 1095610: Additional validations for region/zone names

abhgupta:
- an alphanumeric character anywhere in the name is now required
- there is no regex validation for zone name during deletion

Essentially, we'll perform validations of new region/zone creations but not prevent users from deleting any existing region/zones that were in violation of these new constraints.

### DIFF
--- a/controller/app/models/region.rb
+++ b/controller/app/models/region.rb
@@ -11,7 +11,7 @@ class Region
   index({:name => 1}, {:unique => true})
   create_indexes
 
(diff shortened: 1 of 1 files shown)


abhgupta:
@tiwillia please review

tiwillia:
@abhgupta LGTM.

tiwillia:
[merge]

abhgupta:
re-[merge]

PR #6214, opened 2015-08-07
Removed Double backslashes in variables

ParthKolekar:
Some env vars come up with double backslashes. Although harmless,
some programmers using path.split('/') will be dangerously perplexed by
them. So the strings which are stored in the paths are removed of the
duplicates.

Also substituted the variable for env_dir.

### DIFF
--- a/cartridges/openshift-origin-cartridge-python/usr/versions/2.7-scl/lib/update-configuration
+++ b/cartridges/openshift-origin-cartridge-python/usr/versions/2.7-scl/lib/update-configuration
@@ -8,19 +8,20 @@ function update-configuration {
     # SCL installs of python 2.7
     sclpath=$(dirname $(scl enable python27 "which python"))
     env_dir="${OPENSHIFT_PYTHON_DIR}/env"
-    echo "$OPENSHIFT_PYTHON_DIR/virtenv/bin:$OPENSHIFT_PYTHON_DIR/bin:${sclpath}" > $OPENSHIFT_PYTHON_DIR/env/OPENSHIFT_PYTHON_PATH_ELEMENT
+    echo "${OPENSHIFT_PYTHON_DIR}virtenv/bin:${OPENSHIFT_PYTHON_DIR}bin:${sclpath}" > ${env_dir}/OPENSHIFT_PYTHON_PATH_ELEMENT
 
     local ld_path=$(LD_LIBRARY_PATH="" scl enable python27 "printenv LD_LIBRARY_PATH")
     set_env_var 'OPENSHIFT_PYTHON_LD_LIBRARY_PATH_ELEMENT' $ld_path $env_dir
 
     local man_path=$(MANPATH="" scl enable python27 "printenv MANPATH")
-    path_append ${MANPATH:-:} ${man_path:-:} > $OPENSHIFT_PYTHON_DIR/env/MANPATH
+    path_append ${MANPATH:-:} ${man_path:-:} > ${env_dir}/MANPATH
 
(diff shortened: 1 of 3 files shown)


tiwillia:
[test]

tiwillia:
Test failure is unrelated.

@ParthKolekar Thank you for re-submitting this. Looks good to me now. @abhgupta could you give this a quick review before I merge?

cartridges/openshift-origin-cartridge-python/usr/versions/2.7-scl/lib/update-configuration
```diff
@@ -8,19 +8,20 @@ function update-configuration {
     # SCL installs of python 2.7
     sclpath=$(dirname $(scl enable python27 "which python"))
     env_dir="${OPENSHIFT_PYTHON_DIR}/env"
```
abhgupta: env_dir itself has a slash in there after OPENSHIFT_PYTHON_DIR - doesn't this need to be removed as well?

cartridges/openshift-origin-cartridge-python/usr/versions/shared/lib/update-configuration
```diff
@@ -4,6 +4,7 @@
 
 function update-configuration {
     # Native installations of python
-    echo "$OPENSHIFT_PYTHON_DIR/virtenv/bin:$OPENSHIFT_PYTHON_DIR/bin" > $OPENSHIFT_PYTHON_DIR/env/OPENSHIFT_PYTHON_PATH_ELEMENT
-    echo "$OPENSHIFT_PYTHON_DIR/virtenv" > $OPENSHIFT_PYTHON_DIR/env/VIRTUAL_ENV
+    echo "${OPENSHIFT_PYTHON_DIR}virtenv/bin:${OPENSHIFT_PYTHON_DIR}bin" > ${OPENSHIFT_PYTHON_DIR}/env/OPENSHIFT_PYTHON_PATH_ELEMENT
+    echo "${OPENSHIFT_PYTHON_DIR}virtenv" > ${OPENSHIFT_PYTHON_DIR}/env/VIRTUAL_ENV
```
abhgupta: Did we just miss converting this to env_dir or is this intentional?

cartridges/openshift-origin-cartridge-python/usr/versions/3.3-scl/lib/update-configuration
```diff
@@ -8,19 +8,20 @@ function update-configuration {
     # SCL installs of python 3.3
     sclpath=$(dirname $(scl enable python33 "which python"))
     env_dir="${OPENSHIFT_PYTHON_DIR}/env"
```
abhgupta: Same comment as above.

cartridges/openshift-origin-cartridge-python/usr/versions/2.7-scl/lib/update-configuration
tiwillia: @abhgupta `${env_dir}` is not used in the value of any environment variables, but only to point to the location of the environment variable. Double slashes should be acceptable as long as they aren't in the value of the env variable.

ParthKolekar: It is only a local variable. It dies out soon enough that I felt that it does not matter.

abhgupta:
@tiwillia LGTM

tiwillia:
Thanks! [merge]

PR #6216, opened 2015-08-10
Improve error reporting for member add/remove/update through oo-admin-ctl-domain

tiwillia:
When an error occurs in `domain.save`, an exception is not returned. Instead, the method does not return true to indicate a failure. These failures were not being caught previously.

Additionally, errors encountered in `domain.run_jobs` did not populate `domain.errors`, resulting in an empty hash being reported when an error does occur.

This should resolve both issues.

### DIFF
--- a/broker-util/oo-admin-ctl-domain
+++ b/broker-util/oo-admin-ctl-domain
@@ -335,11 +335,14 @@ when "add_member"
   role = "admin" if role.nil?
   domain.add_members(user, Role.for(role))
   begin
-    domain.save
+    unless domain.save
+      puts "An error occurred during domain modification. Errors: #{domain.errors.messages}"
+      exit 3
+    end
     domain.run_jobs
-  rescue
+  rescue Exception => e
     puts "Unable to add \"#{user.login}\" to domain \"#{domain.canonical_namespace}\" with role \"#{role}\"."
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

tiwillia:
@abhgupta could you review this please? I'd like to get this into 2.2.7

broker-util/oo-admin-ctl-domain
```diff
@@ -335,11 +335,14 @@ when "add_member"
...
-    domain.save
+    unless domain.save
+      puts "An error occurred during domain modification. Errors: #{domain.errors.messages}"
+      exit 3
+    end
     domain.run_jobs
-  rescue
+  rescue Exception => e
     puts "Unable to add \"#{user.login}\" to domain \"#{domain.canonical_namespace}\" with role \"#{role}\"."
-    puts "Errors: #{domain.errors.messages}"
```
abhgupta: how about we check if #{domain.errors.messages} is empty and only then print "Error: #{e.message}" ?

abhgupta:
Other than the comment that applies in 3 places, LGTM

tiwillia:
@abhgupta made the recommended change. Should be ready for merge. PTAL

abhgupta:
[merge] LGTM

PR #6217, opened 2015-08-13
Adding Logic to allow for mvm mirror to be overwritten

sferich888:
There is no way to override the mirror, if you are openshift online.

### DIFF
--- a/cartridges/openshift-origin-cartridge-jbossas/bin/control
+++ b/cartridges/openshift-origin-cartridge-jbossas/bin/control
@@ -207,11 +207,12 @@ function ismgmtup() {
 function build() {
   CONFIG_DIR="${OPENSHIFT_JBOSSAS_DIR}/standalone/configuration"
   OPENSHIFT_MAVEN_MIRROR="${CONFIG_DIR}/settings.base.xml"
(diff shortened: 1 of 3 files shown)


tiwillia:
LGTM, but I am no maven expert. @bparees could you review? Any problems with allowing users to overwrite the `OPENSHIFT_MAVEN_MIRROR` variable?

bparees:
seems ok to me, can you make the same change in our jbossAS and ews cartridges?  (you can do it all in the same commit+PR)

a13m:
@sferich888 , can we talk about the use case here?  We populate a maven mirror and make people use it for a reason.  Bypassing it will typically cost us more and make their performance worse.  Why do we want to allow that?

sferich888:
@a13m I agree allowing for this might be a bad idea. However what if there are outages with the mirror (where customers want a workaround). With out this we have to wait on OpenShift Online to fix the mirrors. This PR is simply to give support a workaround (temporary) for those situations.

sferich888:
All, I am going to close this because of @a13m suggestions. Modifying the correct settings.xml file to point to a different registry is a workaround to these situations. 

All this would do is introduce another layer to the issue.

PR #6218, opened 2015-08-19
Bumping cartridge versions

abhgupta:
@danmcp @bparees Can you please review? Also, it seems to have picked up some changes in addition to the version update.

### DIFF
--- a/cartridges/openshift-origin-cartridge-jbosseap/metadata/manifest.yml
+++ b/cartridges/openshift-origin-cartridge-jbosseap/metadata/manifest.yml
@@ -9,7 +9,7 @@ Version: '6'
 License: LGPL
 License-Url: http://www.gnu.org/copyleft/lesser.txt
 Vendor: Red Hat
-Cartridge-Version: 0.0.24
+Cartridge-Version: 0.0.25
 Compatible-Versions:
 - 0.0.18
 - 0.0.19
@@ -17,6 +17,7 @@ Compatible-Versions:
 - 0.0.21
 - 0.0.22
 - 0.0.23
+- 0.0.24
 Cartridge-Vendor: redhat
 Categories:
 - service
@@ -68,7 +69,7 @@ Endpoints:
     Backend: ''
     Options:
       websocket: true
-  - Frontend: /health
+  - Frontend: "/health"
     Backend: ''
     Options:
       health: true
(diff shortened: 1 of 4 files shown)


cartridges/openshift-origin-cartridge-python/metadata/manifest.yml
```diff
@@ -2,8 +2,8 @@
 Name: python
 Cartridge-Short-Name: PYTHON
 Display-Name: Python 2.6
-Description: ! 'Python is a general-purpose, high-level programming language whose
-  design philosophy emphasizes code readability. Popular development frameworks include:
+Description: 'Python is a general-purpose, high-level programming language whose design
+  philosophy emphasizes code readability. Popular development frameworks include:
```
bparees: i'd go ahead and remove the apostrophes, none of the other carts have them anyway.

cartridges/openshift-origin-cartridge-jbosseap/metadata/manifest.yml
```diff
@@ -68,7 +69,7 @@ Endpoints:
     Backend: ''
     Options:
       websocket: true
-  - Frontend: /health
+  - Frontend: "/health"
```
bparees: i don't think these quotes matter either way, i'm not sure why the tool insists on adding them but it may be a path of least resistance to just leave them.

abhgupta:
@danmcp @bparees ok to merge?

bparees:
Remove the apostrophes on the python description?

abhgupta:
@bparees the sponsorship apostrophes were actually removed by the tool during the manifest file regeneration. Am I missing something?

abhgupta:
Ignore the word "sponsorship" in my comment above. Hazards of typing on the phone.

bparees:
the ! was removed (no idea why that was even there in the first place), the 's are still there.  if you look at the other cart manifests you'll see their descriptions are not contained within ' '.

i don't think they'll do any harm, but might as well make it consistent with the other ones.

bparees:
i'm going to merge this since it should be fine and the tests will catch it if it's not fine, and i know you want it in before the AM stage cut.
[merge]

abhgupta:
@bparees Duh! I read "apostrophe", but my mind just kept processing "exclamation", since that was what had changed. Sorry about the confusion and thanks for the merge. I will fix it shortly for consistency.

PR #6219, opened 2015-08-20
Removing oo-broker prefixes for admin commands

abhgupta:


akostadinov:
why is the change? can such changes be synced with QE first in the future?

abhgupta:
@akostadinov point taken on informing QE ahead of time.

akostadinov:
Thank you. In the beginning I thought change is effective immediately.

abhgupta:
@akostadinov It is effective immediately. It affects the devenv as well as origin (upstream) broker. There is nothing that changes for Online.

akostadinov:
strange, @danmcp wrote it will be effective sprint 103... anyway

abhgupta:
@akostadinov not sure about that - but its all there now.

danmcp:
@akostadinov The end of sprint 103 is when it will end up in production.  Which means we have 4 more weeks to get all the testing lined up.

PR #6220, opened 2015-08-20
Bug 1250904 - Allow ssh keys to be replaced with an empty list

tiwillia:
https://bugzilla.redhat.com/show_bug.cgi?id=1250904

This change allows ssh and krb5 keys for a user to be replaced with an empty list of keys when no keys exist for an application.

The broker passes all krb5 and ssh keys for a user in a single list to this node action. If no ssh keys are passed in the list, it is because no ssh keys exist any longer for the application.

### DIFF
--- a/node/lib/openshift-origin-node/model/application_container_ext/environment.rb
+++ b/node/lib/openshift-origin-node/model/application_container_ext/environment.rb
@@ -248,8 +248,10 @@ module OpenShift
           krb5_principals = ssh_keys.select {|k| k['type'] == 'krb5-principal'}
 
           self.class.notify_observers(:before_replace_ssh_keys, self)
-          AuthorizedKeysFile.new(self).replace_keys(authorized_keys) if authorized_keys.count > 0
-          K5login.new(self).replace_principals(krb5_principals) if krb5_principals.count > 0
+          # If an empty ssh_keys list is passed, this will intentionally replace the ssh keys with an empty list
+          # The broker should pass all keys available for the application to this method
+          AuthorizedKeysFile.new(self).replace_keys(authorized_keys)
+          K5login.new(self).replace_principals(krb5_principals)
           self.class.notify_observers(:after_replace_ssh_keys, self)
 
         end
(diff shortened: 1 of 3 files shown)


tiwillia:
@abhgupta could you review? This ended up being simpler than anticipated.

tiwillia:
[test]

node/lib/openshift-origin-node/model/application_container_ext/environment.rb
```diff
@@ -248,8 +248,8 @@ def replace_ssh_keys(ssh_keys)
           krb5_principals = ssh_keys.select {|k| k['type'] == 'krb5-principal'}
 
           self.class.notify_observers(:before_replace_ssh_keys, self)
-          AuthorizedKeysFile.new(self).replace_keys(authorized_keys) if authorized_keys.count > 0
-          K5login.new(self).replace_principals(krb5_principals) if krb5_principals.count > 0
+          AuthorizedKeysFile.new(self).replace_keys(authorized_keys)
+          K5login.new(self).replace_principals(krb5_principals)
```
abhgupta: Please add a comment in here that confirms that cleaning out the ssh keys if the particular lists are empty is actually intentional.

abhgupta: Could we perhaps also add a simple test for it?

abhgupta:
Change looks good to me

abhgupta:
@tiwillia Can you please address the comments when you get a chance?

tiwillia:
@abhgupta could you look this over again? In adding a test for this scenario, I found that the `AuthorizedKeysFile` model unit test was never run in automated tests since its creation back in 2013. This was because the file name did not match `*_test.rb`. I've fixed this, as well as modified the test to work with the `AuthorizedKeysFile` model since it has certainly changed in the last 2 years.

I also added the comment you mentioned.

abhgupta:
@tiwillia LGTM - feel free to merge when the tests pass

tiwillia:
[merge] Test failures appear to be unrelated.

tiwillia:
re-[merge]

Again, failures are unrelated. Timing out tests are seen often.

PR #6224, opened 2015-08-25
Added & to curl command to prevent potential blocking.

ghost:
@mrunalp could you take a look at this? 

Trying to figure out a better way to work around the blocking that this curl can create. Have other ideas if we get discussion going.

### DIFF
--- a/cartridges/openshift-origin-cartridge-haproxy/bin/control
+++ b/cartridges/openshift-origin-cartridge-haproxy/bin/control
@@ -29,7 +29,7 @@ function isrunning() {
 function ping_server_gears() {
     #  Ping the server gears and wake 'em up on startup.
     for geardns in $(web_gears | cut -f 3 -d ','); do
-          [ -z "$geardns" ]  ||  curl "http://$geardns/" > /dev/null 2>&1  ||  :
+          [ -z "$geardns" ]  ||  (curl "http://$geardns/" > /dev/null 2>&1 &) ||  :
     done
 }
 


cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -29,7 +29,7 @@ function isrunning() {
 function ping_server_gears() {
     #  Ping the server gears and wake 'em up on startup.
     for geardns in $(web_gears | cut -f 3 -d ','); do
-          [ -z "$geardns" ]  ||  curl "http://$geardns/" > /dev/null 2>&1  ||  :
```
Miciah: Eek, I think you want something along the lines of `(curl "http://$geardns/" > /dev/null 2>&1 &)` (which will run curl in the background in a subshell) rather than using backticks (which will try to execute the output of the curl command).

erjones: @Miciah should I correct that in my push? I will happily do so, simply asking as I'm new to this whole thing.

erjones: @Miciah that better? Thank you for the feedback!

Miciah: Looks good to me, but I'll defer to mrunalp regarding the general approach.

erjones: Awesome, appreciate the help either way!

tiwillia:
This LGTM as-is and should resolve https://bugzilla.redhat.com/show_bug.cgi?id=1257757

[test]

tiwillia:
re-[test] failures are unrelated timeouts.

tiwillia:
@mrunalp @danmcp any idea if this is a regression? Has haproxy's ping to all servers always taken up to 4 minutes to complete?

mrunalp:
This PR LGTM

tiwillia:
@danmcp can you close this? Should be fixed by https://github.com/openshift/origin-server/pull/6240

PR #6225, opened 2015-08-25
Removing unnecessary quotes added to the description

abhgupta:


bparees:
lgtm [merge]

abhgupta:
Apparently the quotes are required since there is a 'colon' within the description.

Miciah:
Alternatively, we could delete the colon.  I've known at least one English professor who would argue that our use of the colon here is grammatically incorrect.

abhgupta:
@Miciah We can consider doing that the next time we bump up the cartridge version number.

PR #6229, opened 2015-08-31
Bug 1234603: spreading gears for an app evenly across zones

abhgupta:
Logic for spreading gears:
1. Determine how many zones host gears for this app
   - 1a. If there are zones that have no gears for this app, consider these zones and remove the others
   - 1b. If all zones have gears for this app, do not filter out any zones and continue
2. Distribute gears for an app across zones evenly.
   - Pick zones that have the least number of gears for the app (of the current gear)
3. Distribute all gears (across all apps) across zones evenly
   - From the zones selected above, pick the least consumed zones (have max available capacity)
4. Distribute gears for an app across all nodes evenly
   - From the nodes (belonging to the zones selected above), pick the nodes that have the least number of gears for the app (of the current gear)
5. Distribute all gears (across all apps) across nodes evenly
   - From the nodes selected above, pick the least consumed nodes (have max available capacity)

### DIFF
--- a/broker/test/functional/node_selection_test.rb
+++ b/broker/test/functional/node_selection_test.rb
@@ -103,7 +103,7 @@ class NodeSelectionPluginTest < ActiveSupport::TestCase
   test "least preferred node selection" do
     OpenShift::ApplicationContainerProxy.node_selector_plugin = nil
     begin
-      p = OpenShift::ApplicationContainerProxy.find_available(:node_profile => "small", :least_preferred_servers => [@@server_id])
+      p = OpenShift::ApplicationContainerProxy.find_available(:node_profile => "small", :existing_gears_hosting => {@@server_id => 1})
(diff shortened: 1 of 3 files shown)


abhgupta:
@tiwillia PTAL

abhgupta:
[test]

tiwillia:
@abhgupta these changes LGTM.

There is one other comment we'll want to change though: https://github.com/openshift/origin-server/blob/master/controller/lib/openshift/application_container_proxy.rb#L40

abhgupta:
@tiwillia As discussed, I have made some changes to ensure backward compatibility with any node placement plugin implementation. Instead of converting the existing least_preferred_servers array, I have introduced a new attribute which is a hash of nodes to the number of gears hosted on them for the particular application in question.

tiwillia:
[merge]

PR #6231, opened 2015-09-03
Ensure region controller test can be run concurrently alongside other tests

tiwillia:
Previously, this test deleted all existing regions during setup, created a single region, then expected only one region to be listed. This conflicted with other tests, such as the `broker/test/functional/application_test.rb` which create a region, perform some actions, then delete the region. With this change, no existing regions will be deleted and the test will properly check that the created region is listed in the region output.

### DIFF
--- a/broker/test/functional/regions_controller_test.rb
+++ b/broker/test/functional/regions_controller_test.rb
@@ -3,7 +3,6 @@ require 'test_helper'
 class RegionsControllerTest < ActionController::TestCase
 
   def setup
-    Region.delete_all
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

tiwillia:
@abhgupta @dobbymoodge this should resolve the test failures we are seeing with the enterprise-2.2.7 merge (after its been included in the relevant PR). Could one or both of you review?

Miciah:
Looks good to me.  [merge]

PR #6233, opened 2015-09-04
Bug 1260204 - Only write newer timestamps to last_access files

a13m:


Miciah:
Looks reasonable.  [merge]

Miciah:
Failed because of an AWS EC2 timeout.  [merge] again.

PR #6234, opened 2015-09-09
routing-daemon: F5: Fix initialization

Miciah:
`F5IControlRestLoadBalancerModel#rest_request`: Take an option `:wrap_exceptions`, default value `true`, to indicate whether `rest_request` should rescue exceptions and wrap them in `LBModelException` objects.

`F5IControlRestLoadBalancerModel#initialize`: Specify `wrap_exceptions: false` when getting the openshift_application_aliases policy so that `initialize`'s `rescue RestClient::ResourceNotFound` block will be executed as needed.

This commit fixes a defect introduced in commit bb074e81c11d2eae94b3399939407782d1c38280 that prevented the routing-daemon from creating the openshift_application_aliases policy if it did not already exist during the routing-daemon's initialization.

### DIFF
--- a/routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
+++ b/routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
@@ -46,17 +46,22 @@ module OpenShift
         open_timeout: @open_timeout,
         user: @username,
         password: @password,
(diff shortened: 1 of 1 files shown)


Miciah:
[merge]

@tiwillia, we're going to want to cherry-pick this one too.

PR #6238, opened 2015-09-12
Fix render error

liggitt:


console/app/controllers/application_types_controller.rb
```diff
@@ -138,7 +138,8 @@ def estimate
 
     render :inline => gear_increase_indicator(cartridges, scales, application.gear_profile, false, capabilities, usage_rates, owner)
```
joelsmith: I'd like to see the :inline changed to :text here too (and in the other file too) unless we expect that the output of gear_increase_indicator() would ever be ERB that we want to evaluate. From my reading of it, it shouldn't ever be.

liggitt: yup, updated and made sure the indicator still appeared correctly

Miciah:
Looks good.  [merge].

Miciah:
Failure looks spurious (errors in PHP cartridge's manifest).  [merge] again.

PR #6243, opened 2015-09-21
Bug 1264039: Read all lines even if file does not end in '\n' or EOL

tiwillia:
https://bugzilla.redhat.com/show_bug.cgi?id=1264039

Ensure that all lines of a configuration file are read even if the file does not end in a newline character or EOL.

### DIFF
--- a/logshifter/config.go
+++ b/logshifter/config.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bufio"
+	"io"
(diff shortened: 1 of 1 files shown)


tiwillia:
@codificat fyi

@dobbymoodge or @Miciah could you review?

Miciah:
[merge]

PR #6246, opened 2015-09-22
Check for the exsistance of a skel file that could be use to overrite…

sferich888:
… the gitconfig

This should fix https://bugzilla.redhat.com/show_bug.cgi?id=1265423

### DIFF
--- a/node/lib/openshift-origin-node/model/application_repository.rb
+++ b/node/lib/openshift-origin-node/model/application_repository.rb
@@ -263,8 +263,9 @@ module OpenShift
         }
 
         render_file.call(PathUtils.join(@path, 'description'), 0644, GIT_DESCRIPTION)
-        render_file.call(PathUtils.join(@container.container_dir, '.gitconfig'), 0644, GIT_CONFIG)
-
+        if !File.exist?(PathUtils.join(@container.container_dir, '.gitconfig'))
+            render_file.call(PathUtils.join(@container.container_dir, '.gitconfig'), 0644, GIT_CONFIG)
+        end
         render_file.call(PathUtils.join(hooks, 'pre-receive'), 0755, PRE_RECEIVE)
         render_file.call(PathUtils.join(hooks, 'post-receive'), 0755, POST_RECEIVE)
       end


node/lib/openshift-origin-node/model/application_repository.rb
```diff
@@ -262,9 +262,10 @@ def configure
           File.open(f, 'w', m) { |f| f.write(ERB.new(t).result(binding)) }
         }
 
-        render_file.call(PathUtils.join(@path, 'description'), 0644, GIT_DESCRIPTION)
-        render_file.call(PathUtils.join(@container.container_dir, '.gitconfig'), 0644, GIT_CONFIG)
-
+        render_file.call(PathUtils.join(@path, 'description'), 0644, GIT_DESCRIPTION
```
tiwillia: Missing parenthesis at the end here. Looks like it was moved to the line below.

tiwillia:
The `[pack]` section of the hard-coded gitconfig is important and should always be included. It was added in commit 47065697e2d01521ae650a34ee4fa7c22d49fd26 due to the issue described here:
  http://stackoverflow.com/questions/4826639/repack-of-git-repository-fails

Perhaps we can devise a way to append to the gitconfig, rather than completely overwrite it?

sferich888:
@tiwillia I think that can be solved with documentation! The use case for this should be limited to people looking to customize values in the config, often the [pack] settings. 

In short "if providing your own .gitconfig you need to ensure you provide a [pack] section to avoid said issues".

sferich888:
@tiwillia I should have fixed the issue with the parenthesis

tiwillia:
@sferich888 There are 3 commits, you'll need to squash them.

tiwillia:
LGTM as long as we make sure docs are updated with instructions to include the `[pack]` and `[gc]` sections of the original `.gitconfig`.

@Miciah could you give this a quick review and merge? I'd appreciate a second set of eyes on this.

Miciah:
Looks reasonable... but I wonder whether it would be worth adding a test case.

abhgupta:
Are we waiting on test cases to be added for this?

sferich888:
@Miciah @abhgupta what would we be testing?  If a file exists? Shouldent the core of ruby be providing the tests for that function.

abhgupta:
@Miciah this doesn't seem to require any unit tests - go ahead with the merge if there is nothing else blocking this PR

abhgupta:
@sferich888 @tiwillia make sure we track the documentation for this change.

Miciah:

abhgupta:
@sferich888 @tiwillia Was a doc note added for this?

sferich888:
@abhgupta Docs changes are needed, and would be limited to: 

1) The .gitconfig file can now be overwritten by a .gitconfig file provided in the user's git repostiroy
2) The user-provided .gitconfig should contain the following options to ensure the git repository can be cloned with the gear is close to its memory limits:

	[gc]
	  auto=100
	[pack]
	  windowMemory = 10m
	  packSizeLimit = 20m

abhgupta:
@tiwillia @Miciah can one of you reach out to Docs to make sure this is captured?

Miciah:
I'm following up in the Bugzilla report. which I'll clone to track the required changes to documentation.

PR #6247, opened 2015-09-23
REPORT_BUILD_ANALYTICS=false by default

Miciah:
Change the default value of `REPORT_BUILD_ANALYTICS` in `node.conf` and in the code that uses this setting from `true` to `false` to avoid spurious requests to /broker/analytics.

This commit fixes bug 1111501.
## 

[test]
## 

@abhgupta, is there any reason this would be a bad idea? This change should not impact existing deployments because we have `%config(noreplace) /etc/openshift/node.conf` in `rubygem-openshift-origin-node.spec`, and anyway there probably isn't even anyone using /broker/analytics outside of OpenShift Online.  As for Online, @dak1n1 is making the change to set explicitly the value needed there.

An alternative to making this change here would be to change the installation tools to set REPORT_BUILD_ANALYTICS=false after the fact, but this change seems cleaner.

### DIFF
--- a/node/conf/node.conf
+++ b/node/conf/node.conf
@@ -72,7 +72,7 @@ LIBVIRT_PRIVATE_IP_GW=172.16.0.1
 CONTAINERIZATION_PLUGIN=openshift-origin-container-selinux
 QUOTA_WARNING_PERCENT=90.0
 
-REPORT_BUILD_ANALYTICS=true
+REPORT_BUILD_ANALYTICS=false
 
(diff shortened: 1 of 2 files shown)


dak1n1:
Online is all set, with 'REPORT_BUILD_ANALYTICS=true' merged to master.

Miciah:
Failures in RHC Features look unrelated.  [test] again.

Given that Online is specifying `REPORT_BUILD_ANALYTICS=true` explicitly now, I believe we can merge this change if tests pass.

Miciah:
[merge] please!

PR #6250, opened 2015-09-30
[TO_STAGE] Bug 1260204 - Only write newer timestamps to last_access files

a13m:


a13m:
@dak1n1 , please help me get this into a hotfix release.  Thanks.

dak1n1:
[merge]

wshearn:
[merge]

abhgupta:
re-[merge] - failure seems to have been a flake. instance validation failed before running tests.

tiwillia:
The devenv acceptance script is passing, but also printing additional information:
```
Devenv Acceptance Output = Full path required for exclude: net:[4026532289].
Full path required for exclude: net:[4026532355].
PASS
```

These messages are selinux warnings, probably reported from oo-acept-node running `restorecon`:
```
    do_fail('invalid selinux labels in OPENSHIFT_DIR $GEAR_BASE_DIR') unless %x[/sbin/restorecon -n -v #{$GEAR_BASE_DIR}] =~ //
```

I can't reproduce it on a normal devenv, but only on a stage devenv:
```
# /sbin/restorecon -n -v #{$GEAR_BASE_DIR}
Full path required for exclude: net:[4026532289].
Full path required for exclude: net:[4026532355].
# echo $?
0
```

Something about the selinux policies is different in stage than in master. I'll dig more into this tomorrow.

tiwillia:
It looks like the issue is that this mount exists in the stage devenv, but not in a devenv created from master:
```
# cat /proc/mounts | grep net:
proc net:[4026532290] proc rw,relatime 0 0
proc net:[4026532356] proc rw,relatime 0 0
```

Selinux goes through each mount in `/proc/mounts` and expects the mount point to be a directory. If it isn't, we get the warning message that is causing the devenv to fail acceptance testing.

I've never seen this kind of mount before, and on proc. @a13m @wshearn @abhgupta does anyone know why this exists?

Miciah:
Because this PR has been held up by unrelated issues in the build system, I'm closing it, and we will get the fix in the next release rather than trying to get this in a hot fix.

Miciah:
Oh, I don't have permission to close.  @a13m, would you mind closing it?

PR #6241, opened 2015-09-15
[TO_STAGE] Fix render error

joelsmith:

PR #6252, opened 2015-10-02
oo-accept-node Report errors if selinux contexts on $GEAR_BASE_DIR are incorrect

tiwillia:
Previously, the modified line never reported an error since `~= //` matches anything. A failure will now be reported when restorecon returns any line including 'restorecon reset', indicating that restorecon would have reset the context of $GEAR_BASE_DIR to the proper context if the '-n' flag was not passed.

This change will also cause STDERR from the restorecon command to not be included in the output of `oo-accept-node`, which should just print "PASS" when no failures occur.

Finally, the value of `$GEAR_BASE_DIR` will be printed in the error output rather than the string "$GEAR_BASE_DIR".

### DIFF
--- a/node-util/sbin/oo-accept-node
+++ b/node-util/sbin/oo-accept-node
@@ -355,7 +355,7 @@ def check_selinux()
         do_fail("selinux boolean #{name} should be #{value}") unless result =~ /#{name} --> #{value}/
     }
 
-    do_fail('invalid selinux labels in OPENSHIFT_DIR $GEAR_BASE_DIR') unless %x[/sbin/restorecon -n -v #{$GEAR_BASE_DIR}] =~ //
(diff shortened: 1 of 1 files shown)


tiwillia:
@a13m could you give this a look to make sure it won't affect anything negatively in online?

@abhgupta this should also resolve the stage devenv failures in https://github.com/openshift/origin-server/pull/6250 and https://github.com/openshift/origin-server/pull/6241 when pushed to stage.

tiwillia:
[test]

tiwillia:
@Miciah another example of the net-ssh rhc build failure here.

Miciah:
@tiwillia, yup, it seems to be blocking anything that doesn't fail earlier.

Miciah:
[test] again now that https://www.github.com/openshift/rhc/pull/703 is merged.

Miciah:
Mergey [merge]y.

PR #6253, opened 2015-10-02
Fix typo for gear_whitelist (oo-admin-upgrade)

IshentRas:
Fix output regarding gear_whitelist when using oo-admin-upgrade upgrade-node.

### DIFF
--- a/broker-util/oo-admin-upgrade
+++ b/broker-util/oo-admin-upgrade
@@ -491,8 +491,10 @@ class Upgrader
 
       gears.each do |gear|
         if gear_whitelist && gear_whitelist.length > 0
-          puts "Gear #{gear[:uuid]} is not in the whitelist and will be skipped"
-          next unless gear_whitelist.include?(gear[:uuid])
+          unless gear_whitelist.include?(gear[:uuid])
+            puts "Gear #{gear[:uuid]} is not in the whitelist and will be skipped"
+            next
+          end
         end
 
         if active_gears_map.include?(server_identity) && active_gears_map[server_identity].include?(gear[:uuid])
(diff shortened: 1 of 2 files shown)


broker-util/oo-admin-upgrade
```diff
@@ -490,7 +490,7 @@ class Upgrader
       inactive_gears = []
 
       gears.each do |gear|
-        if gear_whitelist && gear_whitelist.length > 0
+        if !gear_whitelist.include?(gear[:uuid]) && gear_whitelist.length > 0
           puts "Gear #{gear[:uuid]} is not in the whitelist and will be skipped"
           next unless gear_whitelist.include?(gear[:uuid])
```
Miciah: Thanks! There does appear to be a logic error here, but I think we should be leaving the `if` condition the same (because we need to check whether `gear_whitelist` is `nil`) and changing its body from

```ruby
           puts "Gear #{gear[:uuid]} is not in the whitelist and will be skipped"
           next unless gear_whitelist.include?(gear[:uuid])
```

to


```ruby
           unless gear_whitelist.include?(gear[:uuid])
             puts "Gear #{gear[:uuid]} is not in the whitelist and will be skipped"
             next
           end
```

(because we should only print the "will be skipped" message if we actually skip it with `next`).

Does that make sense to you, and if so, would you mind updating the PR please?

IshentRas:
@Miciah : Done. Thanks for the comment. Indeed it does make sense to leave the first condition :+1:

broker-util/oo-admin-upgrade
IshentRas: Just updated the PR

Miciah:
Looks good to me.  Thanks!

[merge]

PR #6254, opened 2015-10-05
[TO_STAGE] oo-accept-node Report errors if selinux contexts on $GEAR_BASE_DIR are incorrect

tiwillia:
Previously, the modified line never reported an error since `~= //` matches anything. A failure will now be reported when restorecon returns any line including 'restorecon reset', indicating that restorecon would have reset the context of $GEAR_BASE_DIR to the proper context if the '-n' flag was not passed.

This change will also cause STDERR from the restorecon command to not be included in the output of `oo-accept-node`, which should just print "PASS" when no failures occur.

Finally, the value of `$GEAR_BASE_DIR` will be printed in the error output rather than the string "$GEAR_BASE_DIR".

This should also resolve the stage devenv failures in #6250 and #6241.

### DIFF
--- a/node-util/sbin/oo-accept-node
+++ b/node-util/sbin/oo-accept-node
@@ -355,7 +355,7 @@ def check_selinux()
         do_fail("selinux boolean #{name} should be #{value}") unless result =~ /#{name} --> #{value}/
     }
 
-    do_fail('invalid selinux labels in OPENSHIFT_DIR $GEAR_BASE_DIR') unless %x[/sbin/restorecon -n -v #{$GEAR_BASE_DIR}] =~ //
+    do_fail("invalid selinux labels in $GEAR_BASE_DIR #{$GEAR_BASE_DIR}") if %x[/sbin/restorecon -n -v #{$GEAR_BASE_DIR} 2>&1].include?("restorecon reset")
 
     # This will likely only detect the effects of a full relabel but it's
     # better than nothing.  My unscientific tests shows that checking 100 gears


tiwillia:
[test]

tiwillia:
Looks like the existing stage devenv issue is causing problems in places other than just the devenv build:
```
  2) Failure:
test_run_as_with_stderr(OpenShift::UtilsSpawnFunctionalTest) [/root/openshift-test/node/test/node_functional/shell_exec_func_test.rb:77]:
<"Good bye, Cruel World\n"> expected but was
<"Full path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\nFull path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\nGood bye, Cruel World\n">.

  3) Failure:
test_run_as(OpenShift::UtilsSpawnFunctionalTest) [/root/openshift-test/node/test/node_functional/shell_exec_func_test.rb:49]:
Expected "Full path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\nFull path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\n" to be empty.

  4) Failure:
test_run_as_env(OpenShift::UtilsSpawnFunctionalTest) [/root/openshift-test/node/test/node_functional/shell_exec_func_test.rb:182]:
Expected "Full path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\nFull path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\n" to be empty.

  5) Failure:
test_run_as_stdout(OpenShift::UtilsSpawnFunctionalTest) [/root/openshift-test/node/test/node_functional/shell_exec_func_test.rb:97]:
Expected "Full path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\nFull path required for exclude: net:[4026532289].\nFull path required for exclude: net:[4026532355].\n" to be empty.
```

Miciah:
I say verify that the fix works in a devenv (if you have not already done so) and petition for a manual merge.

tiwillia:
@Miciah Thanks. The fix here does work, it now successfully builds the devenv. 

However, there are further tests that are failing for the same cause. This change should be merged in for sure, but we'll have to figure out why this issue is occuring on stage devenvs now and never before.

Miciah:
Sorry for the noise—I realised right after posting I had misunderstood... Yeah, so we need to fix other places that are reading the list of mounts and getting confused by `net:` mounts too.

Miciah:
Some guesses as to why the issue is occurring now even though it was not earlier:

• We ran the tests in a container, not on the host, so they didn't see the `net:` mounts (unlikely).

• Something changed in how Docker manages network namespaces so those `net:` mounts weren't there before.

• We used to stop the containers before we ran `rhc-accept-devenv` before—seemed plausible, but these errors from tests would seem to invalidate this hypothesis because the containers have to be running for the tests to run (right?).

• These `net:` mounts weren't there before and are appearing now because of a change in the kernel that affects `/proc/mounts` or a change on the host configuration that affects which file `restorecon` reads—have we checked whether something like `/etc/mtab` changed to a symlink, or the kernel version changed?

• `restorecon` changed—I think we checked that the `policycoreutils` version hadn't changed, right?

tiwillia:
Thanks for the notes @Miciah 

I managed to find an older devenv (ami-2b12e740) where the issue is not occurring. The mount causing the issue is not present in /proc/mounts, so this looks like an update to the way docker handles network namespaces.

Working docker-io version:  docker-io-1.5.0-1.el6.x86_64
Newer docker-io version:     docker-io-1.7.1-2.el6.x86_64

The changelog mostly shows "pulling in upstream", so is not very helpful. Working on finding what has changed. Perhaps there is a way to limit the version of docker used for the devenv.

tiwillia:
Closing this, since there will no longer be a hotfix and this change will be picked up in the next rebase.

tiwillia:
Closing this, since there will no longer be a hotfix and this change will be picked up in the next rebase.

@abhgupta fyi

PR #6256, opened 2015-10-06
routing-daemon: F5: Fix policy if broken

Miciah:
#### routing-daemon: F5: Use const `POLICY_NAME`

Add `OpenShift::F5IControlRestLoadBalancerModel::POLICY_NAME` with value "openshift_application_aliases" and use it instead of literals.
#### routing-daemon: F5: Fix policy if broken

`OpenShift::F5IControlRestLoadBalancerModel#Initialize`: If the policy exists, make sure it has the "forwarding" control and requires the "http" aspect, and patch it if it does not.

This commit fixes bug 1241675.


tiwillia:
@Miciah Looks great to me.

Miciah:
Thanks! [merge].

PR #6258, opened 2015-10-06
routing-daemon: F5: allow multiple hosts

Miciah:
routing-daemon: F5: Support multiple hosts

`F5IControlRestLoadBalancerModel#read_config`: Allow `BIGIP_HOST` to be set to a comma-delimited list of hosts.

`F5IControlRestLoadBalancerModel#rest_request`: Retry the request with successive hosts from the `BIGIP_HOST` setting until either we have tried each host or we get a response.

This commit fixes bug 1227472.
## 

This commit contains changes that will conflict with https://www.github.com/openshift/origin-server/pull/6256, so I have based it on the same branch.


Miciah:
[merge]

PR #6259, opened 2015-10-07
controller: execute_connections: nix client output

Miciah:
Filter client output out from the publish hooks output before providing it as input to subscribe hooks.

This commit fixes bug 1261540.
## 

RFC @dobbymoodge.
## 

[test]


Miciah:
I'm closing this PR because the change will be incorporated into https://www.github.com/openshift/origin-server/pull/6262.

PR #6260, opened 2015-10-08
Fix verbose option for oo-restorecon man page

nak3:


nak3:
Note that since `sh ./rel-eng/misc/convert_man_pages.sh` changed many files, I modified `node-util/man8/oo-restorecon.8` by hand,

tiwillia:
@nak3, you can run `txt2man` without using the script in `rel-eng` to generate the man page for a single *.txt2man file:
`# txt2man -t oo-restorecon -s 8 oo-restorecon.txt2man > oo-restorecon.8`

This LGTM as is, since the only difference when using txt2man would be that the generation timestamp is updated. 

[merge]

nak3:
@tiwillia Oh, I see! Thank you so much!!

PR #6261, opened 2015-10-08
Enhancement at startup of gears and restart of applications (cgroup)

IshentRas:
cg_template_boosted was originally created for handling application creation events (Which require extra resources to start) for instance JEE apps.
This PR is an enhancement for handling applications restart or gears start/restart actions as they also require the same boost.
Seen at customer's side (PM if you require more info)

### DIFF
--- a/node/lib/openshift-origin-node/model/v2_cart_model.rb
+++ b/node/lib/openshift-origin-node/model/v2_cart_model.rb
@@ -1510,7 +1510,9 @@ module OpenShift
             next if options[:secondary_only] and cartridge.name == primary_cartridge.name
             next if options[:exclude_web_proxy] and cartridge.web_proxy?
 
-            buffer << start_cartridge('start', cartridge, options)
+            ::OpenShift::Runtime::Utils::Cgroups.new(@container.uuid).boost do
+              buffer << start_cartridge('start', cartridge, options)
+            end
           end
         else
           buffer << start_gear(options.merge({secondary_only: true}))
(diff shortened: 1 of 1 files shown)


Miciah:
Looks reasonable to me, although I could be missing some reason we were not already boosting these operations.

@ironcladlou, I think you might have some familiarity with the boost code, right? Also, you authored https://github.com/openshift/origin-server/commit/8bee9dbe15cf839b929bc19e9d6eb034f0fb0b05, which changed `unidle_gear` to use a REST call to the broker instead of a boosted direct call to `start_gear`, and this PR would add back the boost in that case, as well as in some other cases.  Would you be able to review this PR?

ironcladlou:
I can't think right now offhand why restart couldn't also be boosted. As far as unidling via the REST API, I remember there was a very good reason that it's necessary, but I'm struggling to remember the details. There might be an old bugzilla issue about it. @jwhonce, I know you and I devised the unidling fix way back... do you happen to remember any more detail?

Miciah:
Yeah, your commit links to https://bugzilla.redhat.com/show_bug.cgi?id=1093776 "Idler doesn't unidle secondary gears in a scaled application".  Your change makes sense, but I think it had the side effect of removing boost for unidle; I thought it was worth mentioning the commit because this PR restores that boost.

jwhonce:
@Miciah @ironcladlou 

History:  V1 only idled the first gear in an application.  That was ported to V2 but Online found a lot of scaled applications where node capacity could be increased if we idled all the gears in the application. Ie, a scaled database application that was created but never used.  Pre @ironcladlou 's PR the database gear never idled.  To idle and then unidle the whole application required orchestration from the Broker.  

A potential issue with this PR is boosting a boosted gear during gear create and updates.  The two test cases I see are that: 1) boosting a boosted gear is restored to default after the outer-most boosted block is finished, and 2) that the gear doesn't loose boost after the embedded boost block is finished.

IshentRas:
Thanks all for the comments so far. As far as I can tell, it has been a real issue at the client's side (starting,restarting JEE application) and I testing the below has definitely boosted the platform.
On over 400 deployed applications on the OSEv2 platform they have got over 80% with JEE cartridges.
They tried to change the resource_limits.conf so as to play with cg_template_throttled or cpu_cfs_quota_us but nothing has worked a part the P.R.

PR #6262, opened 2015-10-08
Bug 1261540 and other pub/sub hooks cleanup

dobbymoodge:
Fix subscriber hooks in haproxy and jboss\* carts to perform better input filtering before parsing.


dobbymoodge:
[test][extended:cartridge]

PR #6263, opened 2015-10-08
haproxy/bin/control: Fix error for no haproxy.cfg

Miciah:
Fix the error message that the haproxy cartridge prints if does not find `haproxy.cfg`.  The error message was referring to the `$f` variable that was deleted in commit f5986a7bddaac5588ab330384e38b1c977ec36b2.
## 

[merge]


Miciah:
EC2 timeouts.  Try to [merge] again.

PR #6264, opened 2015-10-10
oo-diagnostics: Handle empty gemdirs

Miciah:
In `test_for_nonrpm_rubygems`, glob in Ruby rather than in shell in order to avoid getting the glob itself back if it does not match anything.

This commit fixes bug 1101973.


Miciah:
[merge]

PR #6265, opened 2015-10-11
oo-admin-ctl-district-elaborate-on-node-identity

Miciah:
Fix usage information to tell the reader that a node's server identity is its hostname.  Also, mention that regex matching is supported, and fix a typo in the `-r`/`--region` flag description.

This commit fixes bug 1140766.


Miciah:
[merge]

PR #6267, opened 2015-10-12
Set the Host header appropriately in Nginx alias configuration

tiwillia:
Bug 1248439
https://bugzilla.redhat.com/show_bug.cgi?id=1248439

When using nginx as a routing-daemon, the host header should be set to the host provided by the request to the nginx load balancer. Otherwise, the pool_xxx host may be passed as the host header.


tiwillia:
@Miciah would you be able to review this please?

tiwillia:
[test], although I don't believe there are any tests for this component.

Miciah:
Looks good.  Thanks!

[merge] please!

PR #6268, opened 2015-10-13
Kill haproxy during restart even when pid does not exist

tiwillia:
Bug 1211526
https://bugzilla.redhat.com/show_bug.cgi?id=1211526

The haproxy process was not killed with pkill as expected due to a logging message being reported instead of returning non-zero at the end of the `_stop_haproxy_service` function. `pkill` is now run if the haproxy process still exists after the stop during a restart.

The `_stop_haproxy_ctld` method was setting $pid to the haproxy_ctld pid. This same variable is set in `_stop_haproxy_service` only if the haproxy pid file does exist. If the haproxy pid file doesn't exist, the $pid variable still contains the haproxy_ctld pid, causing us to assume that the haproxy pid file does exist.

The stop function would report that the haproxy instance is stopped when the haproxy pid files does not exist. Now, a helpful message will be reported suggesting to run force-stop, as the orignal code appears to have intended.

 Lastly, I modified several `echo` statements to use the sdk's `client_*` methods. This ensures that the user perfoming these actions using rhc will receive the message in the rhc output.

### DIFF
--- a/cartridges/openshift-origin-cartridge-haproxy/bin/control
+++ b/cartridges/openshift-origin-cartridge-haproxy/bin/control
@@ -41,9 +41,9 @@ function wait_to_start() {
 
 function _stop_haproxy_ctld() {
   if [ -f ${HAPROXY_CTLD_PID} ]; then
-    pid=$( /bin/cat "${HAPROXY_CTLD_PID}" )
-    if ps --no-headers --pid ${pid} > /dev/null 2>&1; then
-      kill -TERM $pid
+    local ctld_pid=$( /bin/cat "${HAPROXY_CTLD_PID}" )
+    if ps --no-headers --pid ${ctld_pid} > /dev/null 2>&1; then
+      kill -TERM $ctld_pid
     else
       rm -f ${HAPROXY_CTLD_PID}
     fi
@@ -147,15 +147,19 @@ function _stop_haproxy_service() {
     else
         if `pgrep -x haproxy > /dev/null 2>&1`
         then
-            echo "Warning: HAProxy process exists without a pid file.  Use force-stop to kill." 1>&2
+            client_message "Warning: HAProxy process exists without a pid file."
         else
-            echo "HAProxy already stopped" 1>&2
+            client_message "HAProxy already stopped"
         fi
     fi
 }
 
 function _restart_haproxy_service() {
-    _stop_haproxy_service || pkill haproxy || :
+    _stop_haproxy_service
+    if `pgrep -x haproxy > /dev/null 2>&1`; then
+      client_message "Could not stop HAProxy. Forcefully killing the process."
+      pkill haproxy
+    fi
     _start_haproxy_service
 }
 
@@ -175,17 +179,31 @@ function _reload_service() {
 
 function start() {
     _start_haproxy_service
-    isrunning  &&  echo "HAProxy instance is started"
+    if isrunning; then
+      client_result "HAProxy instance is started"
+    else
+      client_error "HAProxy instance could not be started"
+      exit 1
+    fi
 }
 
 function stop() {
     _stop_haproxy_service
-    isrunning  ||  echo "HAProxy instance is stopped"
+    if `pgrep -x haproxy > /dev/null 2>&1`; then
+      client_message "Could not stop HAProxy. Forcefully killing the process."
+      pkill haproxy
+      if `pgrep -x haproxy > /dev/null 2>&1`; then
+        client_error "Could not stop the HAProxy process."
+        exit 1
+      fi
+    fi
+
+    client_result "HAProxy instance is stopped"
 }
 
 function restart() {
     _restart_haproxy_service
-    isrunning  &&  echo "Restarted HAProxy instance"
+    isrunning  &&  client_result "Restarted HAProxy instance"
 }
 
 function reload() {
@@ -196,14 +214,14 @@ function reload() {
        _reload_service
     fi
 
-    isrunning  &&  echo "Reloaded HAProxy instance"
+    isrunning  &&  client_result "Reloaded HAProxy instance"
 }
 
 function force-reload() {
     if isrunning; then
-        echo "`date`: Conditionally reloading HAProxy service " 1>&2
+        client_message "`date`: Conditionally reloading HAProxy service "
         _reload_service
-        isrunning  &&  echo "Conditionally reloaded HAProxy"
+        isrunning  &&  client_result "Conditionally reloaded HAProxy"
     fi
 }
 


tiwillia:
[test]

Miciah:
Sounds like maybe we should be using `local` more.

tiwillia:
@Miciah do you think it is necessary to use local instead? I don't mind making the change.

Miciah:
Now `_stop` will error out unceremoniously if there is no pid file because it calls `_stop_haproxy_service` and we have `set -e`.

Miciah:
I don't feel too strongly about it, but I think adding `local` would be a good idea while we're modifying this code.  If you do that, you can keep using the name `pid` or rename it to `ctld_pid` in `_stop_haproxy_ctld`, either way.

tiwillia:
@Miciah I made quite a few changes, could you re-review?

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -175,17 +179,22 @@ function _reload_service() {
...
     _start_haproxy_service
-    isrunning  &&  echo "HAProxy instance is started"
+    isrunning  &&  client_result "HAProxy instance is started"
 }
 
 function stop() {
     _stop_haproxy_service
-    isrunning  ||  echo "HAProxy instance is stopped"
+    if `pgrep -x haproxy > /dev/null 2>&1`; then
+      client_error "Could not stop haproxy, use restart or application force-stop to kill."
```
Miciah: Run-on sentence.

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -175,17 +179,22 @@ function _reload_service() {
...
+      exit 1
```
Miciah: Do we want `exit` or `return` here?

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -175,17 +179,22 @@ function _reload_service() {
...
+    else
+      isrunning  || client_result "HAProxy instance is stopped"
```
Miciah: It seems like `isrunning` does approximately the same check as you are doing above: You do `pgrep -x haproxy` above whereas `isrunning` does `ps --pid $haproxy_pid` and then `pgrep -x haproxy -u $current_user`.  It's not clear to me why you cannot just use `isrunning` in `stop()`.

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -196,14 +205,14 @@ function reload() {
        _reload_service
     fi
 
-    isrunning  &&  echo "Reloaded HAProxy instance"
+    isrunning  &&  client_result "Reloaded HAProxy instance"
 }
 
 function force-reload() {
     if isrunning; then
         echo "`date`: Conditionally reloading HAProxy service " 1>&2
```
Miciah: Maybe use `client_message` here?

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -147,15 +147,19 @@ function _stop_haproxy_service() {
     else
         if `pgrep -x haproxy > /dev/null 2>&1`
         then
-            echo "Warning: HAProxy process exists without a pid file.  Use force-stop to kill." 1>&2
+            client_message "Warning: HAProxy process exists without a pid file."
```
Miciah: I wonder whether we should use `client_warn` here and for other warnings.

Miciah: Oh, there isn't a `client_warn` (there is a `warning`, but it doesn't prepend "Warning:" or "CLIENT_" anything).

tiwillia: Right. The closest thing we have is `client_debug`, but I don't think that is appropriate here.

cartridges/openshift-origin-cartridge-haproxy/bin/control
tiwillia: Other cartridges use exit (such as https://github.com/openshift/origin-server/blob/master/cartridges/openshift-origin-cartridge-ruby/bin/control#L121) so I used exit for consistency. I believe they would result in the same behavior, since this is run under `set -e`

tiwillia:
Should `stop()` just use `pkill haproxy` when `_stop_haproxy_service` does not stop the haproxy service? We have already been doing this in `restart()` since 2013. I don't see why it would hurt and we could avoid failing `stop()` with the not-real-helpful error message. 

Additionally, when `_stop_haproxy_service` fails to stop the service, `_stop_haproxy_ctld` has already run, putting the gear in a broken state until it is restarted.

@mrunalp can you comment on the above?

cartridges/openshift-origin-cartridge-haproxy/bin/control
Miciah: Fair enough, yeah, and the only caller of `stop` is the `case` statement at the end, so in the interest of consistency, I suppose this is fine.

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -175,17 +179,22 @@ function _reload_service() {
 
 function start() {
     _start_haproxy_service
-    isrunning  &&  echo "HAProxy instance is started"
+    isrunning  &&  client_result "HAProxy instance is started"
```
Miciah: I feel like this should be an `if` statement that prints a warning if `isrunning` fails, but that may be beyond scope for this PR.

tiwillia: Makes sense. Could `client_error` and `exit 1` otherwise.

mrunalp:
@tiwillia Yeah, I don't see downsides.

cartridges/openshift-origin-cartridge-haproxy/bin/control
tiwillia: `isrunning` returns 1 if the haproxy pid file doesn't exist. If the pid file doesn't exist (due to some race condition) and haproxy is still running, `isrunning` will still return 1. It isn't an accurate check. Perhaps `isrunning` should be modified.

cartridges/openshift-origin-cartridge-haproxy/bin/control
```diff
@@ -147,15 +147,19 @@ function _stop_haproxy_service() {
...
         else
-            echo "HAProxy already stopped" 1>&2
+            client_message "HAProxy already stopped"
         fi
     fi
 }
 
 function _restart_haproxy_service() {
-    _stop_haproxy_service || pkill haproxy || :
+    _stop_haproxy_service || :
```
Miciah: I don't think we need the `|| :` with the `pkill haproxy` command gone.

Miciah:
[merge], please!

PR #6271, opened 2015-10-13
controller: execute_connections: nix client output

Miciah:
Filter client output out from the publish hooks output and chomp it before providing it as input to the subscribe hooks.

This commit is related to bug 1261540.
## 

@dobbymoodge


Miciah:
[merge] please!

PR #6272, opened 2015-10-13
controller: execute_connections: reformat, comment

Miciah:
Reformat `execute_connections` to return right away if the application is not scalable rather than wrapping the entire function in an if block.

Reformat the block that calls the publish hooks to go to the next hook right away if the current hook did not run successfully rather than wrapping the entire block in an `if` block.

Add some comments.
## 

@dobbymoodge


Miciah:
[merge] please!

PR #6275, opened 2015-10-15
oo-accept-node: Test BROKER_HOST & CLOUD_DOMAIN are consistent

dinhxuanvu:
The oo-diagnostics doesn't compare SSL certs from broker & BROKER_HOST
to see if they match since it only collects SSL cert from broker only.

After the commit, the oo-diagnostics gets SSL cert from BROKER_HOST along
with current broker. Then, it compares issuer, subject and common name
records from SSL certs to see if they match. If not, errors are issued.

This commit fixes bug 1054441

### DIFF
--- a/common/bin/oo-diagnostics
+++ b/common/bin/oo-diagnostics
@@ -194,7 +194,7 @@ class OODiag
     @project_is[:origin] = !(@project_is[:enterprise] || @project_is[:online])
     # detect whether the ruby193 SCL is in play.
     system "scl enable ruby193 echo >& /dev/null"
-    @scl_prefix = $?.success? ? "ruby193-" : "" 
+    @scl_prefix = $?.success? ? "ruby193-" : ""
     # detect whether we use service or systemctl
     @use_systemctl = executable? "systemctl"
 
@@ -561,7 +561,7 @@ class OODiag
     a_profile_for.each do |node,profile|
       # check that all nodes are in a district
       if !d_profile_for[node]
-        do_warn <<-NODIST 
+        do_warn <<-NODIST
           Node host #{node} with profile '#{profile}' is not a member of any district.
           Please add it to a district with oo-admin-ctl-district.
         NODIST
@@ -931,7 +931,7 @@ class OODiag
      if have_system_config_firewall
        do_warn <<-WARN
          Using system-config-firewall and lokkit with OpenShift is not recommended.
-         To continue using lokkit please ensure the following custom rules are 
+         To continue using lokkit please ensure the following custom rules are
          installed in #{conf_file}:
 
          --custom-rules=ipv4:filter:/etc/openshift/system-config-firewall-compat
@@ -994,9 +994,9 @@ class OODiag
     if bname && nname && bname != nname
       # broker and node both installed, node may be stealing traffic
       do_warn <<-CONFLICT
-        /etc/httpd/conf.d/000001_openshift_origin_node_servername.conf defines a 
+        /etc/httpd/conf.d/000001_openshift_origin_node_servername.conf defines a
         ServerName #{nname} which may cause the node intercept requests by that name
-        intended for the broker. Please reconfigure with the same ServerName as the 
+        intended for the broker. Please reconfigure with the same ServerName as the
         one in /etc/httpd/conf.d/000002_openshift_origin_broker_servername.conf
       CONFLICT
     end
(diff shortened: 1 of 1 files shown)


node-util/sbin/oo-accept-node
```diff
@@ -337,6 +337,23 @@ def check_node_public_resolution
...
+def check_node_env
+  verbose("checking node env variables")
+
+  broker_host = $CONF.get("BROKER_HOST")
+  cloud_domain = $CONF.get("CLOUD_DOMAIN")
+
+  # check the configuration variables (BROKER_HOST & CLOUD_DOMAIN) against env variables
+  begin
+    broker_host_env = File.open("#{$CONF_DIR}/env/OPENSHIFT_BROKER_HOST", "r").read
+    cloud_domain_env = File.open("#{$CONF_DIR}/env/OPENSHIFT_CLOUD_DOMAIN", "r").read
```
Miciah: Better to use `File.read` because it's shorter and simpler and will close the FD for you.

tiwillia:
The file descriptors need to be closed, or `File.read` should be used as Miciah suggested.

Otherwise, LGTM.

dinhxuanvu:
I fixed the File.open to use File.read instead as suggested. Also, I have added the commit for oo-diagnostics fix.

tiwillia:
[test]

common/bin/oo-diagnostics
```diff
@@ -1226,6 +1228,28 @@ class OODiag
...
     end
 
+    # Retrieve the SSL cert from broker_host_env
+    env_response = `curl -k -s -v https://#{broker_host_env.chomp}/ -o /dev/null 2>&1`
+    if $? != 0
+      do_fail <<-CANNOTCONNECT
+        An error arose while trying to connect to the SSL reverse proxy.  curl returned with
+        #{env_response.strip.empty? ? "no output." : "the following output:\n" + env_response}
+
+        Please check whether the httpd service is running.
```
tiwillia: In this error output, perhaps show that we attempted to connect to `#{broker_host_env}` and failed. Otherwise, the errors for `curl`ing localhost and `#{broker_host_env}` are the same.

common/bin/oo-diagnostics
```diff
@@ -1226,6 +1228,28 @@ class OODiag
...
+
+    # Get ssl cert's issuer, subject and commonname
+    env_issuer = env_response.slice(/^\*\s*issuer: .*$/).gsub(/^\*\s*issuer: /,"")
+    env_subject = env_response.slice(/^\*\s*subject: .*$/).gsub(/^\*\s*subject: /,"")
+    env_commonname = env_response.slice(/^\*\s*common name: .*$/).gsub(/^\*\s*common name: /,"")
+
+    # does ssl cert from broker_host_env matching the current broker
+    do_fail("SSL cert issuer doesn't match between current broker and broker host") if (env_issuer != issuer)
+    do_fail("SSL cert subject doesn't match between current broker and broker host") if (env_subject != subject)
+    do_fail("SSL cert commonname doesn't match between current broker and broker host") if (env_commonname != commonname)
```
tiwillia: These error messages are likely not very helpful to an administrator. The admin would not know the difference between 'current broker' and 'broker host'.

Perhaps something along the lines of:
```
do_fail("SSL cert issuer doesn't match between localhost and #{broker_host_env} defined in /etc/openshift/env/OPENSHIFT_BROKER_HOST")
```

dinhxuanvu:
@tiwillia Fixed the error messages to be more specific as requested.

common/bin/oo-diagnostics
```diff
@@ -1207,14 +1207,16 @@ class OODiag
...
     # Retrieve the SSL cert from Apache
     response = `curl -k -s -v https://localhost/ -o /dev/null 2>&1`
     if $? != 0
       do_fail <<-CANNOTCONNECT
-        An error arose while trying to connect to the SSL reverse proxy.  curl returned with
+        Attempted to connect to #{broker_host_env} but failed. curl returned with
         #{response.strip.empty? ? "no output." : "the following output:\n" + response}
 
-        Please check whether the httpd service is running.
+        Please check whether the httpd service in #{broker_host_env} is running
```
tiwillia: Looks like you made the change in the output for localhost, but meant to make the change in the output for `#{broker_host_env}`

dinhxuanvu:
@tiwillia Fixed the mistake. Sorry about that!

tiwillia:
LGTM assuming tests pass.

common/bin/oo-diagnostics
```diff
@@ -1226,6 +1228,28 @@ class OODiag
...
+        Attempt to connect to #{broker_host_env} but failed. curl returned with
+        #{env_response.strip.empty? ? "no output." : "the following output:\n" + env_response}
+
+        Please check whether the httpd service in #{broker_host_env} is running.
+      CANNOTCONNECT
+      return
+    end
+
+    # Get ssl cert's issuer, subject and commonname
+    env_issuer = env_response.slice(/^\*\s*issuer: .*$/).gsub(/^\*\s*issuer: /,"")
```
Miciah: Might be more elegant to use `env_response.scan(/^\*\s*issuer: (.*)$/).first` or `/^\*\s*issuer: (.*)$/.match(env_response).first`.

common/bin/oo-diagnostics
```diff
@@ -1226,6 +1228,28 @@ class OODiag
...
+    env_subject = env_response.slice(/^\*\s*subject: .*$/).gsub(/^\*\s*subject: /,"")
+    env_commonname = env_response.slice(/^\*\s*common name: .*$/).gsub(/^\*\s*common name: /,"")
+
+    # does ssl cert from broker_host_env matching the current broker
```
Miciah: s/matching/match/

It's good to use full sentences with capitalisation and punctuation, but we don't always do that, so I guess it's not a big deal.

common/bin/oo-diagnostics
```diff
@@ -1226,6 +1228,28 @@ class OODiag
...
+    do_fail("SSL cert issuer doesn't match between localhost and #{broker_host_env} defined in /etc/openshift/env/OPENSHIFT_BROKER_HOST") if (env_issuer != issuer)
```
Miciah: You don't need the parentheses in the `if` condition (or in the `do_fail` call, for that matter), but they don't hurt anything.

common/bin/oo-diagnostics
```diff
@@ -1207,6 +1207,8 @@ class OODiag
   def test_broker_certificate
     skip_test unless @is_broker
 
+    broker_host_env = File.read("/etc/openshift/env/OPENSHIFT_BROKER_HOST")
```
Miciah: Might as well move this closer to where it's used.

common/bin/oo-diagnostics
Miciah: Oh, I see we already use `.slice(//).gsub(//)` above, so never mind.

dinhxuanvu:
@tiwillia @Miciah Fixed minor issues from Miciah.

dinhxuanvu:
@tiwillia @Miciah The testing is good. If you guys are not having any more concerns, please go ahead and merge whenever you can. Thanks.

dobbymoodge:
maybe [merge] ?

dinhxuanvu:
@tiwillia @dobbymoodge @Miciah If there is no other concern, please merge it when you can. Thanks.

Miciah:
[merge] please!

Miciah:
Looks like the merge attempt was foiled by EC2 timeouts.  Please try [merge]ing again!

PR #6278, opened 2015-10-19
oo-admin-ctl-app: Fix remove-gear ignores min scale setting

dinhxuanvu:
Currently, the minimum gear scale limit is ignored when the command
'oo-admin-ctl-app -c remove-gear' is executed causing the application
becomes unscaled when number of gears reaches 1.

The commit will fix a check to compare the number of gears with the
gear limit  which is retrieved from group_overrides. If the number
of gear reaches gear limit, the gear cannot be removed and an error
will be issued. Otherwise, the gear is removed as usual.

Bug <1272195>
Link https://bugzilla.redhat.com/show_bug.cgi?id=1272195

### DIFF
--- a/broker-util/oo-admin-ctl-app
+++ b/broker-util/oo-admin-ctl-app
@@ -30,7 +30,7 @@ Options:
     Gear uuid to operate on
 --cartridge
     Cartridge to operate on
---alias 
+--alias
     Alias to operate on
 --multiplier
     Multiplier factor to be set for the cartridge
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia @Miciah  @dobbymoodge Check it out when you have a chance! No rush though.

tiwillia:
[test]

dinhxuanvu:
@tiwillia @dobbymoodge @Miciah Everything looks good. Please merge it when you can. Thanks.

tiwillia:
@dinhxuanvu looking back over this, I think there may be an issue.

It is possible for a group override to not have the 'min_gears' parameter defined. For instance, when you add cartridge storage to a cartridge for a scaled application that does not have a minimum scaling defined yet, a group override is created that returns `nil` for when `min_gears` is called upon it:
```
irb(main):001:0> a = Application.all.first.group_overrides.first
=> #<GroupOverride @components=[#<ComponentSpec @name="web_proxy" @cartridge_name="haproxy-1.4" @id="562701b9d8542c1dd3000020" @path="haproxy-1.4/web_proxy">, #<ComponentSpec @name="ruby-1.9" @cartridge_name="ruby-1.9" @id="562701b9d8542c1dd300000f" @path="ruby-1.9/ruby-1.9">] @additional_filesystem_gb=1>
irb(main):002:0> a.min_gears
=> nil
```

Comparing nil to an integer with `<=` will cause an exception to be raised.
```

**Comment by dinhxuanvu** [comment]
@tiwillia Fixed the issue as you pointed out. Please re[test].

**Comment by dinhxuanvu** [comment]
@tiwillia Everything is looking good. Please take a look when you have a chance.

File: broker-util/oo-admin-ctl-app [hunk 1]
```diff
@@ -217,6 +217,15 @@ begin
       else
         cis = g.component_instances
         cis.each do |ci|
+          app.group_overrides.each do |ago|
+            unless ago.nil?
+              if !ago.min_gears.nil? && ci.gears.length <= ago.min_gears
+                puts "Gear #{gear_uuid} cannot be removed as the application " +
+                "already reaches the minimum gears limit which is #{ago.min_gears}."
```
tiwillia: Do you mind if we change the wording here a bit? It does get the point across, but perhaps stating it something like the below will be a bit more clear:
```
"has already reached the minimum gear limit of #{ago.min_gears}."
```

dinhxuanvu:
@tiwillia Fixed and re-[test]

tiwillia:
@dinhxuanvu thanks!

[merge]

PR #6282, opened 2015-10-21
haproxy: use POSIX locale for validating endpoints

dobbymoodge:
Bug 1261147
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1261147

Different locales have different ideas of what `[a-z]` mean in a
regex. The regex used for endpoint validation in `update-cluster`
depends on the `POSIX` locale being set. Users that want to control the
locale their apps run in by setting `LC_ALL` to some non-`POSIX`
locale via `rhc set-env LC_ALL=xyz` (e.g. Estonian - `ee_ET`) would find
that scaling events might fail because `[a-z]` in the new locale
excludes some letters that might appear in a valid hostname.

This change updates the `update-cluster` script in the `haproxy`
cartridge so that it preserves the user's locale setting and switches to
the `POSIX` locale only for the part of the script that does the
endpoint validation.

See also:
- https://bugzilla.redhat.com/show_bug.cgi?id=1056666

### DIFF
--- a/cartridges/openshift-origin-cartridge-haproxy/usr/bin/update-cluster
+++ b/cartridges/openshift-origin-cartridge-haproxy/usr/bin/update-cluster
@@ -25,7 +25,8 @@ do
     esac
 done
 
-
+ORIG_LC_ALL="${LC_ALL}"
+REGEX_LC_ALL="POSIX"
 #[ $# -gt 3 ] || print_help
 
 exitcode=0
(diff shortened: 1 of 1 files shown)


dobbymoodge:
[test][extended:gear]

dobbymoodge:
@bparees

bparees:
lgtm.

Miciah:
[merge] please!

PR #6283, opened 2015-10-21
Bug 1064039: Add oo-diagnostic report 401 Unauthorized error

dinhxuanvu:
The 'oo-diagnostics' currently won't fail if there are 410 Unauthorized
error in the production/development.log. After this change, 'oo-diagnostics'
will check either production or development.log depending on which mode the
broker is in. If the 401 errors exist, an error message will be displayed
with the sugguested solution to fix the issue. The warning message may
continue to be displayed in the subsequential 'oo-diagnostics'
until the 401 error message is cleared up in the next log rotation.

Bug <1064039>
Link https://bugzilla.redhat.com/show_bug.cgi?id=1064039

### DIFF
--- a/common/bin/oo-diagnostics
+++ b/common/bin/oo-diagnostics
@@ -194,7 +194,7 @@ class OODiag
     @project_is[:origin] = !(@project_is[:enterprise] || @project_is[:online])
     # detect whether the ruby193 SCL is in play.
     system "scl enable ruby193 echo >& /dev/null"
-    @scl_prefix = $?.success? ? "ruby193-" : "" 
+    @scl_prefix = $?.success? ? "ruby193-" : ""
     # detect whether we use service or systemctl
     @use_systemctl = executable? "systemctl"
 
@@ -561,7 +561,7 @@ class OODiag
     a_profile_for.each do |node,profile|
       # check that all nodes are in a district
       if !d_profile_for[node]
-        do_warn <<-NODIST 
+        do_warn <<-NODIST
           Node host #{node} with profile '#{profile}' is not a member of any district.
           Please add it to a district with oo-admin-ctl-district.
         NODIST
@@ -931,7 +931,7 @@ class OODiag
      if have_system_config_firewall
        do_warn <<-WARN
          Using system-config-firewall and lokkit with OpenShift is not recommended.
-         To continue using lokkit please ensure the following custom rules are 
+         To continue using lokkit please ensure the following custom rules are
          installed in #{conf_file}:
 
          --custom-rules=ipv4:filter:/etc/openshift/system-config-firewall-compat
@@ -994,9 +994,9 @@ class OODiag
     if bname && nname && bname != nname
       # broker and node both installed, node may be stealing traffic
       do_warn <<-CONFLICT
-        /etc/httpd/conf.d/000001_openshift_origin_node_servername.conf defines a 
+        /etc/httpd/conf.d/000001_openshift_origin_node_servername.conf defines a
         ServerName #{nname} which may cause the node intercept requests by that name
-        intended for the broker. Please reconfigure with the same ServerName as the 
+        intended for the broker. Please reconfigure with the same ServerName as the
         one in /etc/httpd/conf.d/000002_openshift_origin_broker_servername.conf
       CONFLICT
     end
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia @dobbymoodge @Miciah Please take a look when you have a chance. Thanks

common/bin/oo-diagnostics
```diff
@@ -1204,6 +1204,29 @@ class OODiag
 
   end
 
+  def test_auth_keys_salt_failure
+    if Rails.env.development?
+      msg_log = `grep \'401 Unauthorized\' /var/log/openshift/broker/development.log`
+    else
+      msg_log = `grep \'401 Unauthorized\' /var/log/openshift/broker/production.log`
+    end
```
Miciah: Would be better to use a variable for the log filename and interpolate that variable into the command.

common/bin/oo-diagnostics
```diff
@@ -1204,6 +1204,29 @@ class OODiag
 
   end
 
+  def test_auth_keys_salt_failure
+    if Rails.env.development?
+      msg_log = `grep \'401 Unauthorized\' /var/log/openshift/broker/development.log`
```
Miciah: You don't need to escape the single-quotes in backticks.

common/bin/oo-diagnostics
```diff
@@ -1204,6 +1204,29 @@ class OODiag
...
+    else
+      msg_log = `grep \'401 Unauthorized\' /var/log/openshift/broker/production.log`
+    end
+
+    unless msg_log.empty?
+      do_fail <<-CONFLICT
+      Detect an '401 Unauthorized' error on the log. This error indicates the
```
Miciah: Using the past tense ("Detected") would make this error message clearer.

Miciah: Maybe better phrased something like, "This error may indicate inconsistent authentication keys or salts among brokers".

common/bin/oo-diagnostics
Miciah: Is this going to cause errors for 401s generated by regular clients?

dinhxuanvu:
@Miciah Fixed several wording suggestions that you noticed. The code now checks for a bit more specific error than "401 Unauthorized". Unfortunately, the error log for this bug doesn't have enough info to verify if this is from user-end or dev-end.

dinhxuanvu:
[test]

dinhxuanvu:
re[test]

dinhxuanvu:
@tiwillia

tiwillia:
@dinhxuanvu oo-diagnostics can be run on either a node or broker. The test you have written is specific only to brokers and should not be run on nodes. You need something like this at the beginning of your method: 
```
 skip_test unless @is_broker
```
*From https://github.com/dinhxuanvu/origin-server/blob/oo-diagnostics-sync/common/bin/oo-diagnostics#L668*

common/bin/oo-diagnostics
```diff
@@ -1204,6 +1204,33 @@ class OODiag
...
+      mode = "production"
+      msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
+    end
+
+    unless msg_log.empty?
+      do_warn <<-AUTH_ERROR
+      Detected an '401 Unauthorized' error on the #{mode} log. This error
+      may indicate inconsistent authentication keys or salts among brokers.
+      Please follow below steps to fix the issue:
+      1. Manually synchronize the authentication keys and salt
```
tiwillia: This is rather minor, but could you include " in the broker's configuration" to indicate to the user where the keys need to be synchronized?

common/bin/oo-diagnostics
```diff
@@ -1204,6 +1204,33 @@ class OODiag
...
+  def test_auth_keys_salt_failure
+    log_dir = "/var/log/openshift/broker/"
+
+    if Rails.env.development?
+      mode = "development"
+      msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
+    else
+      mode = "production"
+      msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
+    end
```
tiwillia: There are three rails environments used: "development", "test", and "production". If rails is using the "test" environment, your current code will still look in production.log even though it may not exist or have the relevant log messages. Perhaps a case statement is needed.

dinhxuanvu:
@tiwillia  Fixed and re-[test]

common/bin/oo-diagnostics
```diff
@@ -1204,6 +1204,39 @@ class OODiag
...
+
+    if Rails.env.development?
+      mode = "development"
+      msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
+    elsif Rails.env.production?
+      mode = "production"
+      msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
+    else
+      mode = "test"
+      msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
```
tiwillia: The `msg_log =` line does not need to be repeated three times. Place it only once after the `if...end` statement since it is the same in each case.

dinhxuanvu:
re-[test]

dinhxuanvu:
@tiwillia All things look good. Please double-check and merge if possible. Thanks!

tiwillia:
LGTM, thanks @dinhxuanvu !

[merge]

PR #6285, opened 2015-10-21
FrontendHttpServer: Recover from missing manifest

Miciah:
`OpenShift::Runtime::FrontendHttpServer#initialize`: Rescue any exception from initializing the cartridge model, and set `@standalone_web_proxy` to false in that contingency.

Before this commit, failure to initialize the cartridge model would cause a failure to initialize the frontend http server, which would cause a failure to initialize the container plugin, which would prevent the container plugin's `destroy` method from finishing.  Consequently, it was impossible to delete a gear with a bad `manifest.yml` file.

This commit fixes bug 1273658.
## 

[test] [extended:gear]
## 

@jwhonce, does this look reasonable?

### DIFF
--- a/node/lib/openshift-origin-node/model/frontend_httpd.rb
+++ b/node/lib/openshift-origin-node/model/frontend_httpd.rb
@@ -178,7 +178,7 @@ module OpenShift
         @container_name = container.name
         @namespace = container.namespace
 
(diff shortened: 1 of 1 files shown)


jwhonce:
@Miciah Seems to be a reasonable compromise for these errors.  LGTM.

Miciah:
Thanks!

PR #6287, opened 2015-10-23
Bug 1264722: Fix oo-register-dns shows errors with any option

dinhxuanvu:
There is a command conflict as '-h' option is linked link both '--help'
and '--with-node-hostname' which causes 'getoptlong' object to fail.
In fact, '--help' should be linked to '-?' instead. This commit recrifies
the issue by linking '--help' back to '-?' as it shows in help message.

Bug <1264722>
Link https://bugzilla.redhat.com/show_bug.cgi?id=1264722

### DIFF
--- a/broker-util/oo-register-dns
+++ b/broker-util/oo-register-dns
@@ -1,13 +1,13 @@
 #!/usr/bin/env oo-ruby
 #--
 # Copyright 2010 Red Hat, Inc.
-# 
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #    http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia [test]

dinhxuanvu:
@Miciah @dobbymoodge This would be very easy to check and merge.

Miciah:
The commit that introduced the conflict between the two `-h` flags was commit 5840883fb7d3448f838bdd110c078057498b26dd.  The same commit changed `-?` to `-h` in several other commands, but `oo-register-dns` is the only command where I see a conflict.  Moreover the Bugzilla reports mentioned in that commit do not mention `oo-register-dns`, so I do not believe that reverting back to `-?` in `oo-register-dns` will cause any regressions with regard to those Bugzilla reports.

I wish we could instead change the `-h` short option for `--with-node-hostname`, but that would break existing users of the flag, so doing so is not feasible at this point; changing the flag back to `-?` is probably the best possible solution.

Thanks, Vu! openshift-bot, please [merge]!

PR #6291, opened 2015-10-26
Fix gear extended tests

Miciah:
When appending a line for the `GEAR_UPGRADE_EXTENSION` setting to `/etc/openshift/node.conf`, include a leading `\n` EOL marker in case the last line of `node.conf` is incomplete and is missing the EOL marker.  Otherwise the `GEAR_UPGRADE_EXTENSION` setting will be appended to the incomplete line and will not take effect, and as a consequence, the test will fail the first time it runs (although it will succeed on subsequent runs).

Change the clean-up code to delete the entire line for the `GEAR_UPGRADE_EXTENSION` setting instead of substituting an empty line, and then delete the last line if it is empty.
## 

[test] [extended:gear]

### DIFF
--- a/controller/test/cucumber/step_definitions/upgrade_steps.rb
+++ b/controller/test/cucumber/step_definitions/upgrade_steps.rb
@@ -387,7 +387,7 @@ end
 EOF
 
   IO.write('/tmp/gear_upgrade.rb', gear_upgrade_content)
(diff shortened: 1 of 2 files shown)


tiwillia:
LGTM [merge]

Miciah:
Thanks!

Miciah:
Thanks, openshift-bot!

PR #6292, opened 2015-10-27
Bug 1265609: Fix pandas not getting installed

dinhxuanvu:
Pandas package fails to install properly if it's included in setup.py for
Python 2 & 3 applications. This issue is possibly due to a bug with numpy
dependency of pandas package. This commit allows pandas package to be
installed using 'pip install -e .' instead of 'python setup.py install'
to work around the numpy bug.

Bug <1265609>
Link https://bugzilla.redhat.com/show_bug.cgi?id=1265609

### DIFF
--- a/cartridges/openshift-origin-cartridge-python/usr/versions/shared/bin/control
+++ b/cartridges/openshift-origin-cartridge-python/usr/versions/shared/bin/control
@@ -182,9 +182,17 @@ function build() {
         ( cd $OPENSHIFT_REPO_DIR; pip install -r ${OPENSHIFT_REPO_DIR}${requirements_file} $OPENSHIFT_PYTHON_MIRROR $OPENSHIFT_PIP_TRUSTED_HOST )
     fi
 
+    echo "Checking pip install marker.."
+    export_pip_install
+
     if [ -f ${OPENSHIFT_REPO_DIR}/setup.py ]; then
-        echo "Running setup.py script.."
-        ( cd $OPENSHIFT_REPO_DIR; python ${OPENSHIFT_REPO_DIR}/setup.py develop $OPENSHIFT_PYTHON_MIRROR )
+        if [ $OPENSHIFT_PYTHON_USE_PIP == "enable" ]; then
+          echo "Running pip install.."
+          ( cd $OPENSHIFT_REPO_DIR; pip install -e . )
+        else
+          echo "Running setup.py script.."
+          ( cd $OPENSHIFT_REPO_DIR; python ${OPENSHIFT_REPO_DIR}/setup.py develop $OPENSHIFT_PYTHON_MIRROR )
+        fi
     fi
 
     relative-virtenv
--- a/cartridges/openshift-origin-cartridge-python/usr/versions/shared/lib/util
+++ b/cartridges/openshift-origin-cartridge-python/usr/versions/shared/lib/util
@@ -1,6 +1,14 @@
 #!/bin/bash
 # Utility functions for use in the cartridge scripts.
 
+function export_pip_install() {
+  if marker_present "pip_install"; then
+    export OPENSHIFT_PYTHON_USE_PIP="enable"
+  else
+    export OPENSHIFT_PYTHON_USE_PIP="disable"
+  fi
+}
+
 function appserver_pid() {
 	pgrep -f "python -u app.py"
 }


dinhxuanvu:
@tiwillia @Miciah @dobbymoodge Please review when you have a chance. Thanks. [test]

dinhxuanvu:
@soltysh @rhcarvalho Would you guys mind talking a look at this bug and its fix as well? I would really appreciate the consultation on this.

Here are my findings and explanation (excerpt from a sent email) regarding the bug:

I attach the copies (see [A] & [B]) of command execution verbose from both "python setup.py install" and the new "pip install -e .". Also, here are the link [1] to the unknown numpy dependency bug that is probably associated with the pandas bug and also a link [2] to stackoverflow where I found the suggested solution.

[A] http://pastebin.test.redhat.com/323063
[B] http://pastebin.test.redhat.com/323066
[1] https://github.com/numpy/numpy/issues/2434
[2] http://stackoverflow.com/questions/25753276/install-numpy-pandas-as-dependency-in-setup-py
[3] http://python-packaging-user-guide.readthedocs.org/en/latest/pip_easy_install/
[4] http://stackoverflow.com/questions/3220404/why-use-pip-over-easy-install

Essentually, 'pip install' option uses the pip install itself to install three main dependencies (including numpy) of the pandas package (please refer to A & B documents) while 'python setup.py install' is using setuptools (easy_install) I think (see [3]). It's simply better at dependency-resolving and that's all it is as far as I can see. Python install looks like it just goes straight to install pandas package without carefully verifying the dependencies associated it and end up failing. The known bug with numpy listed above is possibly contributing to the error as well.

In my very own point of view, pip install option is safe as I have tested with several popular packages such as django and it is fine. In fact, the pip install option is preferable option as I found several online conversations pointing its advantages (see [4]). We can simply replace the python install with pip install completely if necessary.

Thanks in advance!

cartridges/openshift-origin-cartridge-python/usr/versions/shared/bin/control
```diff
@@ -183,8 +183,15 @@ function build() {
...
-        echo "Running setup.py script.."
-        ( cd $OPENSHIFT_REPO_DIR; python ${OPENSHIFT_REPO_DIR}/setup.py develop $OPENSHIFT_PYTHON_MIRROR )
+        msg=$(grep 'pandas' ${OPENSHIFT_REPO_DIR}/setup.py)
+        if [ -z "$msg" ]
+        then
+          echo "Running setup.py script.."
+          ( cd $OPENSHIFT_REPO_DIR; python ${OPENSHIFT_REPO_DIR}/setup.py develop $OPENSHIFT_PYTHON_MIRROR )
+        else
+          echo "Running pip install.."
+          ( cd $OPENSHIFT_REPO_DIR; pip install -e . )
```
soltysh: You probably want to pass `$OPENSHIFT_PYTHON_MIRROR $OPENSHIFT_PIP_TRUSTED_HOST` like we do that [a couple lines before](https://github.com/openshift/origin-server/pull/6292/files#diff-b02b0980d560b2675bbee4f4748a60e6L182).

soltysh:
One nit, and the change itself LGTM.

As for the transition, from my research, it's perfectly ok to make the transition from `python setup.py develop` to `pip install -e . `. Even more, it's highly encouraged to do that. Additionally we are already using `pip install -r requirements.txt` couple lines before so this will unify the entire install process. 
The only question is do we want such a big change? @danmcp

dinhxuanvu:
@soltysh Thanks for the review and consultation. Certainly appreciated it. I agreed with the moving to pip install since it's newer and more reliable (as you said we already use it). If @danmcp is fine with this, I will just use pip install as default and remove the conditional statement + python install in the code as well.

danmcp:
@soltysh @dinhxuanvu I am only ok with this in general if we are 100% confident in no regressions for existing apps.  It seems like a big change.

dinhxuanvu:
@tiwillia @abhgupta

abhgupta:
@soltysh @danmcp The reason we went down the path of using pip for installing these dependencies was that `python setup.py install` was failing for at least one package. Additionally, we didn't want to make a one-off change in the code that hard-coded the specific package (pandas). 

An alternative approach (not sure if its required or recommended), would be to always use `python setup.py install` first and if it fails then fall back to `pip install`. Does that make sense?

danmcp:
@abhgupta I think I would prefer an environment variable to let people choose to use pip and the default stay the same.

abhgupta:
@danmcp I would prefer that as well. Keep `python setup.py install` as the default that can be overridden by the user env var. Keeping the existing install option as default will ensure we don't break existing applications.

abhgupta:
@dinhxuanvu I believe we have settled on an approach

soltysh:
:+1:

dinhxuanvu:
@abhgupta @tiwillia I have changed to go with the approach that we discussed above. Please review when you can. [test]

cartridges/openshift-origin-cartridge-python/usr/versions/shared/lib/util
```diff
@@ -1,6 +1,14 @@
 #!/bin/bash
 # Utility functions for use in the cartridge scripts.
 
+function export_pip_install() {
+	if marker_present "pip_install"; then
```
soltysh: A nit, but can you fix the indentation - it reads better :wink:

soltysh:
LGTM. Don't forget to update the docs with info about the marker!

dinhxuanvu:
@soltysh Thanks for the comment. Fixed and [test]

abhgupta:
[merge]

dinhxuanvu:
@tiwillia The merge fails and it looks like an EC2 timeout error. Would you mind taking a look to verify that it's not related to my changes and re-merge it if possible? Thanks

PR #6293, opened 2015-10-28
Improving the waitdeployments script to check application endpoints

sferich888:
Bug 1270660
https://bugzilla.redhat.com/show_bug.cgi?id=1270660

I'm not sure I like this as much as I like the original commit because it will change the cartridges behavior compared to what it currently does. 

However with that said the, code now more closely mimics what the EAP cartridge is doing by _at least_ checking one application context (via curl) before returning and saying the cartridge is started.

### DIFF
--- a/cartridges/openshift-origin-cartridge-jbossews/bin/control
+++ b/cartridges/openshift-origin-cartridge-jbossews/bin/control
@@ -87,13 +87,26 @@ function ishttpup() {
     return 1
 }
 
-# This function is here to mimmic the EAP cartriges design: 
-# Due to the lack of a provided mgmt interface, the options to check for 
-#  deployment activation is limited. Thus this fucntion uses an ENV VAR to 
-#  allow you configure a sleep time (in seconds), giving your deployments 
-#  time to start before marking the gear as active. 
+# This function is here to mimmic the EAP cartriges design:
+# Due to the lack of a provided mgmt interface, the options to check for
+#  deployment activation is limited. Thus this fucntion uses an ENV VAR to
+#  allow you configure a sleep time (in seconds), giving your deployments
+#  time to start before marking the gear as active.
 function waitondeployments() {
-    sleep ${OPENSHIFT_JBOSSEWS_START_DELAY:-0}
+    attempts=${OPENSHIFT_JBOSSEWS_START_TRIES:-5}
+    sleep_time=${OPENSHIFT_JBOSSEWS_HEALTH_WAIT_TIME:-2}
+    context=${OPENSHIFT_HAPROXY_HTTPCHK_URI:-/}          # context must start with '/'
+    context=${OPENSHIFT_JBOSSEWS_HEALTH_CONTEXT:-$context}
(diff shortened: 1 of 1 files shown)


sferich888:
[test]

cartridges/openshift-origin-cartridge-jbossews/bin/control
```diff
@@ -87,13 +87,26 @@ function ishttpup() {
...
+# This function is here to mimmic the EAP cartriges design:
+# Due to the lack of a provided mgmt interface, the options to check for
+#  deployment activation is limited. Thus this fucntion uses an ENV VAR to
+#  allow you configure a sleep time (in seconds), giving your deployments
+#  time to start before marking the gear as active.
 function waitondeployments() {
-    sleep ${OPENSHIFT_JBOSSEWS_START_DELAY:-0}
+    attempts=${OPENSHIFT_JBOSSEWS_START_TRYS:-5}
+    sleep_time=${OPENSHIFT_JBOSSEWS_HEALTH_WAIT_TIME:-2}
+    context=${OPENSHIFT_HAPROXY_HTTPCHK_URI:-/}          # context must start with '/'
```
Miciah: `OPENSHIFT_HAPROXY_HTTPCHK_URI` will be present exactly when the framework cartridge is collocated with a proxy, right? And the framework cartridge will be collocated with the proxy on exactly the first gear in non-HA applications, and on some but generally not all subsequent gears with an HA application.

cartridges/openshift-origin-cartridge-jbossews/bin/control
```diff
@@ -87,13 +87,26 @@ function ishttpup() {
...
-#  allow you configure a sleep time (in seconds), giving your deployments 
-#  time to start before marking the gear as active. 
+# This function is here to mimmic the EAP cartriges design:
+# Due to the lack of a provided mgmt interface, the options to check for
+#  deployment activation is limited. Thus this fucntion uses an ENV VAR to
+#  allow you configure a sleep time (in seconds), giving your deployments
+#  time to start before marking the gear as active.
 function waitondeployments() {
-    sleep ${OPENSHIFT_JBOSSEWS_START_DELAY:-0}
+    attempts=${OPENSHIFT_JBOSSEWS_START_TRYS:-5}
```
Miciah: s/TRYS/TRIES/

codificat:
wondering if there should be a "switch" to enable the whole `waitondeployments` check, and leave it disabled by default... to keep the default behaviour as it is right now.
Not sure if adding yet another var to control this would be good... or just set the default for OPENSHIFT_JBOSSEWS_START_TRIES to 0?

tiwillia:
We don't like the fact that this introduces several new tunables and is really just a workaround. But, since there is no mgmt interface exposed by jbossews-2.0, we are not aware of any better solution.

[merge]

tiwillia:
openshift-bot, there should be a stable build of 'libra_ami' now, please [merge].

PR #6297, opened 2015-10-29
Delete explicit haproxy component overrides when disabling ha

tiwillia:
Bug 1255426
https://bugzilla.redhat.com/show_bug.cgi?id=1255426

When an application has a minimum and/or maximum scaling limit set, it is possible that implicit overrides for HA applications will be set as explicit overrides in the database. This previously caused applications that were made ha, then had a scaling limit set, then had ha disabled to be unable to scale back down to a single haproxy cartridge.

With this change, the explicit overrides will be removed if they exist when an HA application has HA disabled.

### DIFF
--- a/broker/test/functional/app_events_controller_test.rb
+++ b/broker/test/functional/app_events_controller_test.rb
@@ -149,10 +149,12 @@ class AppEventsControllerTest < ActionController::TestCase
     post :create, "event" => "disable-ha", "application_id" => @app.id
     assert_response :success
     overrides = @app.reload.group_instances_with_overrides
+    explicit_overrides = @app.group_overrides
     assert !@app.ha
     assert_equal 1, overrides.length
     assert_equal 1, overrides[0].min_gears
     assert_equal(-1, overrides[0].max_gears)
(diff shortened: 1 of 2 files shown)


tiwillia:
[test]

tiwillia:
@abhgupta could you review?

The conditions I use to select the component in the explicit group overrides to delete are somewhat strict. Do you see a need for any changes or additional checks?

An application that has been scaled to a min of 3 after being made HA will have the following explicit overrides in `group_overrides`:
```
=> [#<GroupOverride @components=[#<ComponentOverrideSpec @name="web_proxy" @cartridge_name="haproxy-1.4" @id="562fbbfa1efa3f1757000020" @path="haproxy-1.4/web_proxy" @min_gears=2 @max_gears=-1 @multiplier=0>, #<ComponentSpec @name="ruby-1.9" @cartridge_name="ruby-1.9" @id="562fbbfa1efa3f175700000f" @path="ruby-1.9/ruby-1.9">] @min_gears=3>]
```

abhgupta:
[merge]

PR #6298, opened 2015-10-30
 Add tc support for multiple IFs via TRAFFIC_CONTORL_DEVS

joelsmith:
Add a new configuration item, TRAFFIC_CONTORL_DEVS, which is a
space-delimited list of interfaces that should be tc-managed.

Update code so that tc changes are applied across all devices that
are to be traffic-controlled.

Also hard-code limits to very high numbers for the loopback interface if it is to be managed.


abhgupta:
@Miciah can you please take a look at this PR and provide your review/feedback?

Miciah:

PR #6299, opened 2015-10-30
Bug 1218872: Fix rhc setup fail during upload sshkey

dinhxuanvu:
Uploading ssh key during 'rhc setup' command may fail due to the method
'get_minimum_ssh_key_size' returns nil if MINIMUM_SSH_KEY_SIZE variable
is not set in broker.conf file.

This commit change the '||' operation to use rescue instead in order to
handle the nil case from the first condition and return the default value
'DEFAULT_MINIMUM_SSH_KEY_SIZE' which is zero by default in case nil value
detected.

Bug 1218872
Link https://bugzilla.redhat.com/show_bug.cgi?id=1218872

### DIFF
--- a/controller/app/models/ssh_key.rb
+++ b/controller/app/models/ssh_key.rb
@@ -11,18 +11,18 @@ class SshKey
   include Mongoid::Document
   include AccessControlled
 
-  # This is the current regex for validations for new ssh keys 
+  # This is the current regex for validations for new ssh keys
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia @Miciah Please review when you can. Thanks! [test]

tiwillia:
LGTM [merge]

PR #6303, opened 2015-11-02
Bug 1276460: oo-admin-repair --orphaned-envs no need to  populate hashes

dinhxuanvu:
The 'oo-admin-repair' commands populate hashes for user, domain, app and district
regardless if they are necessary or not. After this change only needed hashes are
for specific commands by using several flags to determine which commands are used.
This change will improve the performance by avoid populating unnecessary hashes.

The current "--ssh-keys" option removes orphaned environment variables which should
only be done by using "--orphaned-envs" option. This change removes that task from
"--ssh-keys" option. Furthermore, with this change, the orphaned variables are now
only removed from the domains but not from the gears as they are updated when the
applications are updated. Even if the orphaned variables remain in the gears, there
are no harms. Also, the orphaned variables will be grouped by domain and then removed
all at once for each domain to reduce database access. All changes are intended to
further improve the overall performance of the oo-admin-repair command.

Bug 1276460
Link https://bugzilla.redhat.com/show_bug.cgi?id=1276460

### DIFF
--- a/broker-util/oo-admin-repair
+++ b/broker-util/oo-admin-repair
@@ -159,7 +159,7 @@ def reset_consumed_gears(user_id)
   rescue Mongoid::Errors::DocumentNotFound
     puts "User #{user_id} not found."
   rescue Exception => e
-    puts "Unable to set consumed gears for user #{user_id}, error: #{e.message}." 
+    puts "Unable to set consumed gears for user #{user_id}, error: #{e.message}."
   end
   return success
 end
@@ -384,7 +384,7 @@ def populate_available_servers
     district_info["server_names"].each do |si|
       available_servers[si] = (district_info['name'] || 'NONE')
     end
-  end 
+  end
 
   $datastore_hash.each do |gear_uuid, gear_info|
     si = gear_info['server_identity']
@@ -408,7 +408,7 @@ def find_unresponsive_servers(available_servers, auto_confirm)
     puts "Checking for unresponsive servers...FAIL" if $verbose
     puts "Servers that are unresponsive:"
     uservers = unresponsive_servers.dup
-    uservers.each do |server| 
+    uservers.each do |server|
       print "\tServer: #{server} (district: #{available_servers[server]}), Confirm [yes/no]: "
       unresponsive_servers.delete(server) unless auto_confirm or (auto_confirm.nil? and gets.chomp.to_b)
       puts ""
@@ -457,7 +457,6 @@ def find_orphaned_envs(namespace=nil)
     # Difference the lists to get all component_id's specified by an env var for components that no longer exist
     missing_cids = cid_list - existing_cid_list
     orphaned_envs = all_envvar_cids.select {|document| missing_cids.include?(document["cid"]) }
-
     orphaned_envs = orphaned_envs.map {|oe| {:domain_id => oe["_id"], :component_id => oe["cid"]} }
   end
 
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
@tiwillia [test]

broker-util/oo-admin-repair
```diff
@@ -952,10 +952,13 @@ $chk_gear_mongo_node = true if fix_ssh_keys or fix_district_uids or fix_removed_
 $chk_district = true if fix_district_uids or fix_removed_nodes
 $chk_usage = true if fix_usage
 
-populate_user_hash
-populate_domain_hash
-populate_district_hash
-populate_app_hash
+# Skip populating hashes if --orphaned-envs is used
+if fix_orphaned_envs.nil?
```
tiwillia: This logic is probably incorrect. Multiple flags can be passed into `oo-admin-repair`. So if I run `oo-admin-repair --consumed-gears --orphaned-envs`, the hashes will not be populated and the command will likely fail.

dinhxuanvu:
@tiwillia Thanks for the pointer. Fixed!

broker-util/oo-admin-repair
```diff
@@ -952,10 +952,15 @@ $chk_gear_mongo_node = true if fix_ssh_keys or fix_district_uids or fix_removed_
...
-populate_app_hash
+# Skip populating hashes if --orphaned-envs is used
+unless fix_ssh_keys.nil? and fix_district_uids.nil? and fix_consumed_gears.nil? and
+   fix_removed_nodes.nil? and fix_usage.nil? and fix_gear_sizes.nil? and
+   !fix_orphaned_envs.nil?
+  populate_user_hash
+  populate_domain_hash
+  populate_district_hash
+  populate_app_hash
+end
```
abhgupta: Not a big fan of this approach. If we add another flag/option to the tool, we'll have to remember to add it here as well. Instead, we should probably set a flag (either one for all data hashes or individually for different hashes) based on what options/flags are being passed. 

For starters, I'd vote for having a single boolean be used to determine whether all the data hashes need to be populated or not.

dinhxuanvu:
@tiwillia I have added a variable as a flag as @abhgupta commented above. [test]

dinhxuanvu:
@abhgupta @tiwillia Please review when you can. Thanks! [test]

broker-util/oo-admin-repair
```diff
@@ -930,15 +923,29 @@ end
...
-    domain = Domain.find(orphan[:domain_id])
+  puts "Removing orphaned environment variables from domain(s)..." if $verbose
+  domain_component = orphaned_envs.group_by{ |d| d[:domain_id] }.each{ |_, v| v.map!{ |d| d[:component_id] }}
+  domain_component.each do |domain_key, component_array|
     begin
-      domain.remove_env_variables(orphan[:component_id])
-      puts "Removed envs with component id #{orphan[:component_id]} from domain \"#{domain.namespace}\""
-    rescue => e
-      print_message "Unable to remove envs with component id #{orphan[:component_id]} from domain \"#{domain.namespace}\": #{e.message}"
+      variables = []
```
tiwillia: This is not a very descriptive name for this variable. Perhaps something like `env_vars = []`.

Calling any variable `variable` is generally a bad idea.

dinhxuanvu:
@tiwillia Thanks for the notice. Changed it to env_vars_to_remove to be more descriptive.

tiwillia:
@dinhxuanvu this LGTM, can't find any other issues with it. Lets get @abhgupta's blessing before we merge.

dinhxuanvu:
@tiwillia Sound good. Thanks.

abhgupta:
[merge]

PR #6308, opened 2015-11-10
Bug 1254521: Fix give proper prompt when deleting region with asterisk

dinhxuanvu:
The oo-admin-ctl-region command crashes when confirm prompt is displayed
for user's input due to line break error.

This commit fixes the issue by removing the line break before displaying
the prompt.

Bug 1254521
Link https://bugzilla.redhat.com/show_bug.cgi?id=1254521

### DIFF
--- a/broker-util/oo-admin-ctl-region
+++ b/broker-util/oo-admin-ctl-region
@@ -45,8 +45,8 @@ opts = GetoptLong.new(
     ["--command",   "-c", GetoptLong::REQUIRED_ARGUMENT],
     ["--region",    "-r", GetoptLong::REQUIRED_ARGUMENT],
     ["--zone",      "-z", GetoptLong::REQUIRED_ARGUMENT],
-    ["--description", "-d", GetoptLong::REQUIRED_ARGUMENT],    
-    ["--bypass",    "-b", GetoptLong::NO_ARGUMENT],    
+    ["--description", "-d", GetoptLong::REQUIRED_ARGUMENT],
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia Pretty simple one to clear :dart:

tiwillia:
[test]

tiwillia:
@dinhxuanvu please re-visit this. I was able to reproduce the issue in the bug. After implementing the fix in this PR, it did not resolve the isssue:
```
# oo-admin-ctl-region -c destroy -r *
!!!! WARNING !!!! WARNING !!!! WARNING !!!!
You are about to delete the devenv-local region.

This is NOT reversible, all remote data for this region will be removed.
Do you want to delete this region (y/n): /usr/sbin/oo-admin-ctl-region:121:in `gets': Is a directory - jenkins (Errno::EISDIR)
	from /usr/sbin/oo-admin-ctl-region:121:in `gets'
	from /usr/sbin/oo-admin-ctl-region:121:in `<main>'
```

dinhxuanvu:
@tiwillia Fixed. Sorry about that. [test]

tiwillia:
LGTM. 

I think `$stdin` is usually preferred over `STDIN`, due to it being a global variable rather than a constant and being used by default in methods like `gets`. However, we use both throughout the code, so I don't think it matters.

tiwillia:
Just a note, the attached bug is silly. They are running the command incorrectly and expecting it to work, because it did with `create`. However, the fix here doesn't hurt anything and actually makes the gets more specific, so it won't read from extra options on the command line. 

[merge]

PR #6310, opened 2015-11-11
[TO_STAGE] Add tc support for multiple IFs via TRAFFIC_CONTORL_DEVS

joelsmith:
Add a new configuration item, TRAFFIC_CONTORL_DEVS, which is a
space-delimited list of interfaces that should be tc-managed.

Update code so that tc changes are applied across all devices that
are to be traffic-controlled.

Also hard-code limits to very high numbers for the loopback interface if it is to be managed.

These commits have already been merged to master in #6298

### DIFF
--- a/controller/test/cucumber/step_definitions/node_steps.rb
+++ b/controller/test/cucumber/step_definitions/node_steps.rb
@@ -326,14 +326,20 @@ end
 
 Then /^a traffic control entry should( not)? exist$/ do |negate|
   acctname = @account['accountname']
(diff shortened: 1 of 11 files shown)


wshearn:
[merge]

joelsmith:
[Test] again, please

Miciah:
@joelsmith, looks like we need https://github.com/openshift/rhc/pull/703 in stage.

Miciah:
Oh, looks like stage is already has that commit.  The solution isn't that simple...

joelsmith:
@Miciah, the master PR, #6298 failed the same tests during its first test run, then was re-tested (automatically?) by the bot over the weekend, and the tests passed that time, so it merged.

PR #6311, opened 2015-11-12
deployments_controller.rb: Adds check for REST deployment without arguments

thrasher-redhat:
deployments_controller.rb: Adds check for REST deployment without arguments

Bug 1223025
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1223025

Currently, attempting to change a deployment and not passing in a target variable (either ref or artifact_url) will result in a false success message.

Adds a check to ensure that a ref or artifact_url variable is present before attempting to deploy.

### DIFF
--- a/broker/test/functional/deployments_controller_test.rb
+++ b/broker/test/functional/deployments_controller_test.rb
@@ -132,4 +132,18 @@ class DeploymentsControllerTest < ActionController::TestCase
     end
   end
 
+  test "check for both binary and git deployment input" do
+    # Check for cases where we have both a ref and an artifact_url
+    test_url = "http://localhost/test.tgz"
+    test_ref = "branchname"
+    post :create, { "application_id" => @app._id, "artifact_url" => test_url, "ref" => test_ref }
+    assert_response :unprocessable_entity, "Cannot specify a git deployment and a binary artifact deployment"
+  end
+
+  test "check for no binary or git deployment input" do
+    # Check for cases where we have have neither a ref nor an artifact_url
+    post :create, { "application_id" => @app._id, "artifact_url" => nil, "ref" => nil }
+    assert_response :unprocessable_entity, "Must specify a git deployment or a binary artifact deployment"
+  end
+
 end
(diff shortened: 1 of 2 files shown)


controller/app/controllers/deployments_controller.rb
```diff
@@ -40,6 +40,9 @@ def create
     return render_error(:unprocessable_entity, "Cannot specify a git deployment and a binary artifact deployment",
                           -1, "artifact_url") if artifact_url && ref
 
+    return render_error(:unprocessable_entity, "Must specify a git deployment or a binary artifact deployment",
+                          -1, "artifact_url") unless artifact_url || ref
```
tiwillia: You pass `"artifact_url"` as the field in the `render_error`. This is the field that the error supposedly applies to. Since it applies to `artifact_url` and `ref`, I would probably leave it blank.

tiwillia:
Other than the one note, this LGTM. We should probably have a test for it.

thrasher-redhat:
Added test for this case (and another similar case that wasn't covered in testing).

tiwillia:
[test]

tiwillia:
LGTM, lets make sure tests pass before we merge.

broker/test/functional/deployments_controller_test.rb
```diff
@@ -132,4 +132,18 @@ def teardown
...
+  test "check for both binary and git deployment input" do
+    # Check for cases where we have both a ref and an artifact_url
+    test_url = "http://localhost/test.tgz"
+    test_ref = "branchname"
+    post :create, { "application_id" => @app._id, "artifact_url" => test_url, "ref" => test_ref }
+    assert_response :unprocessable_entity, "Cannot specify a git deployment and a binary artifact deployment"
+  end
+
+  test "check for no binary or git deployment input" do
+    # Check for cases where we have have neither a ref and an artifact_url
```
Miciah: s/and/nor/

Miciah:
Looks good but for a grammatical error in a comment.

broker/test/functional/deployments_controller_test.rb
thrasher-redhat: Thats what I get for making changes on a Friday...fixing...

tiwillia:
[merge]

tiwillia:
Please try to [merge] again, as the failures appear to be unrelated.

PR #6312, opened 2015-11-15
Bug 1170540 - False success oo-admin-cartridge --recursive/-R in invalid dir

dinhxuanvu:
When a cartridge is installed with recursive option and an invalid directory,
the oo-admin-cartridge command fails to report error and falsely succeed.

This commit adds a invalid/empty directory check to the oo-admin-cartridge
to stop execution and report the error to the user.

Bug 1170540
Link https://bugzilla.redhat.com/show_bug.cgi?id=1170540

### DIFF
--- a/node-util/sbin/oo-admin-cartridge
+++ b/node-util/sbin/oo-admin-cartridge
@@ -36,6 +36,11 @@ module OpenShift
                      [options.source]
                    end
 
+            if dirs.empty?
+              $stderr.puts "Installation failed: Invalid or empty path to cartridge source: '#{options.source}'"
+              raise OpenShift::Runtime::Utils::ShellExecutionException.new("Installation failed", 1, "no cartridges were found.")
+            end
+
             success = true
 
             dirs.each do |dir|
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia @Miciah [test]

node-util/sbin/oo-admin-cartridge
```diff
@@ -36,6 +36,11 @@ module OpenShift
                      [options.source]
                    end
 
+            if dirs.empty?
+              $stderr.puts "Provided directory is either empty or invalid."
+              return "Failed"
```
tiwillia: This works and is a good change, but I think the wording can be improved. Take a look at the same failure when not providing the recursive flag:
```
Installation failed for /usr/libexec/openshift/cartridges/abc: Illegal path to cartridge source: '/usr/libexec/openshift/cartridges/abc'

Operation failed: some cartridges cannot be installed
```

We should be using similar wording. Returning just "Failed" is not very descriptive about the current state of the cartridge. "Operation failed: some cartridges cannot be installed" or something similar would be better.

tiwillia:
LGTM besides the wording of the new error.

dinhxuanvu:
@tiwillia All fixed and [test] please.

tiwillia:
[merge] please!

dinhxuanvu:
@Miciah @tiwillia Some unexpected mongodb error. Can you guys verify it's not related to my code and re-merge it perhaps?

tiwillia:
Odd, that is the same thing that happened on another PR earlier today. Lets see if it happens again [merge]

PR #6314, opened 2015-11-17
Fix component-overrides-after-disable-ha test

Miciah:
Fix the test added in commit 006ac9886ec85dab6639d0668da9e978a5d1bece that makes sure the component override that is added when a scaling limit is set on an HA application is cleared when HA is disabled for that application:
- Split the test into a distinct test for this specific issue.
- Set the scaling minimum (which was a required step to reproduce the problem that commit 006ac9886ec85dab6639d0668da9e978a5d1bece fixed).  Otherwise `@app.group_overrides` will be empty (which is why the test was not failing before this commit).
- Iterate over the components of each group override, rather than merely iterating over the group overrides themselves.
- Fix a typo: `ComponentOverrrideSpec` -> `ComponentOverrideSpec`.
## 

Please [test]!
## 

@tiwillia

### DIFF
--- a/broker/test/functional/app_events_controller_test.rb
+++ b/broker/test/functional/app_events_controller_test.rb
@@ -22,7 +22,8 @@ class AppEventsControllerTest < ActionController::TestCase
     @domain = Domain.new(namespace: @namespace, owner:@user)
     @domain.save
     @app_name = "app#{@random}"
-    @app = Application.create_app(@app_name, cartridge_instances_for(:php), @domain, :scalable => true)
+    @cart = cartridge_instances_for(:php).first
+    @app = Application.create_app(@app_name, [@cart], @domain, :scalable => true)
     @app.save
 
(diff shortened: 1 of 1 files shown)


broker/test/functional/app_events_controller_test.rb
```diff
@@ -149,15 +150,35 @@ def teardown
...
+    assert_response :success
+    application_controller, @controller = @controller, EmbCartController.new
+    put :update, "scales_from" => "3", "scales_to" => "3", "application_id" => @app.id, "id" => @cart.name
+    assert_response :success
+    assert @app.reload.gears.count == 3
+    @controller = application_controller
+    post :create, "event" => "disable-ha", "application_id" => @app.id
+    assert_response :success
+    assert !@app.ha
+    @app.group_overrides.each do |override|
```
tiwillia: Is it not necessary to reload here?

tiwillia:
LGTM, assuming tests pass :+1:

Miciah:
Thanks!

PR #6315, opened 2015-11-18
oo-admin-chk: Adds solutions to a couple of error messages

thrasher-redhat:
Bug 1111598
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1111598

Changes the incorrect error message for when a gear exists on the node, but not in mongo.  The error message now tells the user the correct solution for the error.

Also adds a solution to the error message for when there is a mismatch between consumed gears and actual gears.

### DIFF
--- a/broker-util/oo-admin-chk
+++ b/broker-util/oo-admin-chk
@@ -169,9 +169,11 @@ node_hash.each do |gear_uuid, gear_info|
   datastore_gear_info = $datastore_hash[gear_uuid]
   if !datastore_gear_info
     if !datastore_has_gear?(gear_uuid)
+      server_identity = gear_info[0]
       puts "#{gear_uuid}...FAIL" if $verbose
-      print_message "Gear #{gear_uuid} exists on node #{gear_info[0]} (uid: #{gear_info[1]}) but does not exist in mongo database"
-      print_message "Please see https://access.redhat.com/solutions/1171163 for more information."
+      print_message "Gear #{gear_uuid} exists on node #{server_identity} (uid: #{gear_info[1]}) but does not exist in mongo database"
+      print_message "To fix this issue, remove the gear by running the oo-devel-node command from the node '#{server_identity}':"
+      print_message "oo-devel-node app-destroy --with-container-uuid #{gear_uuid}"
     end
   elsif !datastore_gear_info['gear_uid'].nil?
     begin
--- a/controller/app/helpers/admin_helper.rb
+++ b/controller/app/helpers/admin_helper.rb
@@ -459,6 +459,8 @@ module AdminHelper
         user_consumed_gears, app_actual_gears = check_consumed_gears(user)
         if user_consumed_gears != app_actual_gears
           print_message "User #{owner_hash['login']} has a mismatch in consumed gears (#{user_consumed_gears}) and actual gears (#{app_actual_gears})"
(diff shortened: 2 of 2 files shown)


broker-util/oo-admin-chk
```diff
@@ -171,7 +171,7 @@ node_hash.each do |gear_uuid, gear_info|
     if !datastore_has_gear?(gear_uuid)
       puts "#{gear_uuid}...FAIL" if $verbose
       print_message "Gear #{gear_uuid} exists on node #{gear_info[0]} (uid: #{gear_info[1]}) but does not exist in mongo database"
-      print_message "Please see https://access.redhat.com/solutions/1171163 for more information."
+      print_message "To fix this issue, remove the gear by deleting /var/lib/openshift/#{gear_uuid} and removing the user from the node host."
```
tiwillia: Why not suggest using the `oo-devel-node destroygear` command instead of manually deleting the directory and user?

Miciah: "removing the user from the node host"—does this refer to editing `/etc/passwd`? It might help to be more explicit here.  (You can say "or other configured name service" for the off chance that the customer is using some other NSS configuration.)

broker-util/oo-admin-chk
```diff
@@ -171,7 +171,8 @@ node_hash.each do |gear_uuid, gear_info|
     if !datastore_has_gear?(gear_uuid)
       puts "#{gear_uuid}...FAIL" if $verbose
       print_message "Gear #{gear_uuid} exists on node #{gear_info[0]} (uid: #{gear_info[1]}) but does not exist in mongo database"
-      print_message "Please see https://access.redhat.com/solutions/1171163 for more information."
+      print_message "To fix this issue, remove the gear by using the oo-devel-node command:"
```
tiwillia: One tiny nit-pick, lets be specific about where this command should be run. Users will run oo-admin-chk on a broker, but will need to run oo-devel-node on which node the gear is installed on. If we have the `server_identity` we should point the user there. Otherwise, just let them know it needs to be run on the node.

Miciah: The `oo-admin-chk` command is on the broker, and the `oo-devel-node` command is on the node, right? If the user runs a command on the broker, and the output of that first command just says to run a second command but doesn't say where to run it, the user is likely to try to run the second command on the same host.

broker-util/oo-admin-chk
```diff
@@ -169,9 +169,11 @@ node_hash.each do |gear_uuid, gear_info|
   datastore_gear_info = $datastore_hash[gear_uuid]
   if !datastore_gear_info
     if !datastore_has_gear?(gear_uuid)
+      server_identity = gear_info[0]
       puts "#{gear_uuid}...FAIL" if $verbose
       print_message "Gear #{gear_uuid} exists on node #{gear_info[0]} (uid: #{gear_info[1]}) but does not exist in mongo database"
```
tiwillia: Since you defined `server_identity` as `gear_info[0]`, you should replace `#{gear_info[0]}` here with `server_identity` as to not duplicate code.

thrasher-redhat:
[test]

tiwillia:
LGTM, we'll merge assuming tests pass.

tiwillia:
[merge]

PR #6316, opened 2015-11-18
Bug 1261156 - Fix oo-admin-xx help command returns an exitcode 255 instead of 0

dinhxuanvu:
Currently, many oo commands with help option return an exitcode 255 instead of 0.
This commit will specifically return 0 for --help option and 255 for incorrect
options.

Bug 1261156
Link https://bugzilla.redhat.com/show_bug.cgi?id=1261156

### DIFF
--- a/broker-util/oo-admin-broker-auth
+++ b/broker-util/oo-admin-broker-auth
@@ -48,7 +48,6 @@ Options:
 -h|--help
     Show Usage info
 USAGE
-  exit 255
 end
 
 opts = GetoptLong.new(
@@ -67,11 +66,13 @@ begin
   opts.each{ |k,v| args[k]=v }
 rescue GetoptLong::Error => e
   usage
+  exit 255
 end
 
 # Check this _before_ loading the rails env
 if args["--help"] || (STDIN.tty? && args.empty?)
   usage
+  exit 0
 end
 
 if args['--timeout']
--- a/broker-util/oo-admin-broker-cache
+++ b/broker-util/oo-admin-broker-cache
@@ -42,7 +42,6 @@ Options:
 -h|--help
     Show Usage info
 USAGE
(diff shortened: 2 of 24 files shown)


dinhxuanvu:
@tiwillia @Miciah [test]

dinhxuanvu:
@Miciah @tiwillia This is a low-priority PR but would like to have it reviewed and merged to clear up the log :D. Thanks.

tiwillia:
@dinhxuanvu The changes you've made look good. I ran through a loop of every oo-command and ran them with --help to determine which still did not report a '0' exit code. Those that did not report '0' did not support a --help, -h, or help argument, so the non-zero code is expected.

The oo-gear-registry command was the only command that has a --help option and results in a non-zero exit code. This appears to be due to an error which results in the non-zero exit code:
```
# oo-gear-registry --help
  NAME:

    ["OpenShift Gear Registry Access"]

  DESCRIPTION:

    ["Access utility for gear registry"]

  COMMANDS:
	
    help                 Display global or [command] help documentation.	

  GLOBAL OPTIONS:
	
    -h, --help 
        Display help documentation
	
    -v, --version 
        Display version information
	
    -t, --trace 
        Display backtrace when an error occurs
	
/opt/rh/ruby193/root/usr/share/gems/gems/openshift-origin-node-1.38.4/lib/openshift-origin-node/model/application_container.rb:141:in `from_uuid': container_uuid is required! (ArgumentError)
	from /usr/bin/oo-gear-registry:35:in `<main>'
```
We should probably fix that in another PR, but its out-of-scope for this one.

LGTM. [merge]

dinhxuanvu:
@tiwillia Would you mind re-merging this as the net-ssh PR is already merged.

tiwillia:
[merge]

PR #6322, opened 2015-11-24
man/oo-admin-gear.8: Fixes incorrect flag option

thrasher-redhat:
Bug 1283372
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1283372

The man page for oo-admin-gear incorrectly stated that the short flag for --with-container-uuid was -u.  Attempting to use -u will result in an error.  The actual flag is -c, which is consistent with other oo-admin commands.

Changes man page to show '-c' instead of the incorrect '-u'.


Miciah:
Thanks! openshift-bot, please [merge]!

PR #6324, opened 2015-11-24
Bug 1075420 - 'git not found' when validate scalable app with database via oo-idler-stats.

dinhxuanvu:
Scalable app with database has each cartridge on separate gear and some gears with database
cartridge should not contain git repo. When oo-idler-stats checks those gears, it will fail
to find git repo.

This commit modifies oo-idler-stats to only look for git repo for web_framework cartridge
and ignores git check for other types of cartridges.

This commit also fixes various broken codes such as cartridge_types method does not work
at all and run method also returns a list with empty string due to mis-use split method.

Bug 1075420
Link https://bugzilla.redhat.com/show_bug.cgi?id=1075420

### DIFF
--- a/node-util/sbin/oo-idler-stats
+++ b/node-util/sbin/oo-idler-stats
@@ -106,7 +106,6 @@ def user_list():
             dct[l[1]] = l[0]
     return dct
 
-
 def run(cmd, workdir=None):
     """ run commands in a subprocess and wait for the return code.
         optional: provide a working dir for the command.
@@ -118,7 +117,7 @@ def run(cmd, workdir=None):
 
     proc = Popen(cmd, shell=True, stdin=PIPE, stdout=PIPE, stderr=PIPE)
     output = proc.communicate()
-    output = output[0].split('\n')
+    output = output[0].strip().split('\n')
 
     if cwd:
         os.chdir(cwd)
@@ -131,7 +130,7 @@ def percent(a, b):
 
 
 def cartridge_types(homedir):
-    return_code, types = run("ls |egrep -v '-|git'", homedir)
+    return_code, types = run("ls |egrep -v -e '-|git'", homedir)
     return types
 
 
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia @Miciah Please review and [test]. Thanks!

node-util/sbin/oo-idler-stats
```diff
@@ -131,7 +130,7 @@ def percent(a, b):
 
 
 def cartridge_types(homedir):
-    return_code, types = run("ls |egrep -v '-|git'", homedir)
+    return_code, types = run("ls |egrep -v -e '-|git'", homedir)
```
tiwillia: Could you explain why this change was necessary? The previous syntax looks correct according to the egrep manpage and some quick testing.

dinhxuanvu: @tiwillia That code didn't work as it failed while I was testing it. Apparently, due to the pattern starting with '-' in '-|git', it's recommended to use -e to avoid the egrep misinterpret the pattern as a option. It's mentioned on manpage.

tiwillia: Ah, that does make sense. Thank you.

tiwillia:
@Miciah or @dobbymoodge could one of y'all review? This lgtm, but I'm not extremely familiar with python and do not feel like I would be able to catch most potential issues.

node-util/sbin/oo-idler-stats
```diff
@@ -144,40 +143,65 @@ def get_app_type(app_path):
...
+        # Check if cartridge has been checked for cartridge type before
+        if cart in cartridges_list:
+            # Flag git_repo if this cartridge should have a git repo
+            if cartridge_types.get(cart) == true:
+                git_repo = True
+                break
+        else:
+            cmd = " ".join(["cat",os.path.join(app_path, cart, "metadata/manifest.yml"),
+                                               "| grep web_framework"])
+            ignore, result = run(cmd)
```
dobbymoodge: It's more "pythonic" to use "_" instead of "ignore" for placeholders when unpacking tuples.

dinhxuanvu:
@jwhonce I would like to have your inputs on this PR since it involves your original codes. Thanks!

dinhxuanvu:
@jwhonce I have removed the optimization part to force the method to check yaml files in all cases to avoid the issue with potential false reports on different same-name cartridge. Let me know if you find any other issues. Thanks.

dinhxuanvu:
re-[test]

dinhxuanvu:
re-[test] please.

dinhxuanvu:
@jwhonce I wonder if I have your blessing to merge this?

jwhonce:
LGTM

dinhxuanvu:
@Miciah @tiwillia I have blessing from Jhon, would you mind merging this? Thanks!

node-util/sbin/oo-idler-stats
```diff
@@ -144,40 +143,54 @@ def get_app_type(app_path):
...
-
-    # Is there stuff inside git/appname.git?
-    gitfiles = os.listdir(app_path + "/git/" + app_name + ".git")
-    if len(gitfiles) == 0:
-        return "Path %s/git/%s.git is empty." % (app_path, app_name)
-
-    # Chances are good that this is a git repo. What's the last commit?
-    last_commit = run('git show HEAD | grep Date | sed "s/Date:   //"',
-                      app_path + "/git/" + app_name + ".git")[1][0]
+    global app_name
```
Miciah: Why make `app_name` a global?

dinhxuanvu: app_name is a variable out of the scope of the method. In order to modify it inside the method, it needs to be global. This is for reusing the app name that is collected from the git repo from the first gear for the subsequent gears from the same app which may not have git repo.

Miciah: This means that you are assuming that `validate` will be called first on a gear with a framework cartridge, and then subsequently on any other gears from the same application, correct? Looking at https://github.com/dinhxuanvu/origin-server/blob/idler-stats/node-util/sbin/oo-idler-stats#L231-L247, it is not obvious to me (1) that gears will be grouped by application and (2) that gears will be sorted such that a gear with a framework cartridge comes before related gears without.

dinhxuanvu:
@Miciah [test] & ready for review.

Miciah:
Please [merge]!

PR #6325, opened 2015-11-25
routing-daemon: F5: Fix policy check

Miciah:
`F5IControlRestLoadBalancerModel`#`initialize`: Check if the policy's "controls" or "requires" field is defined before checking whether the field has the correct value.


tiwillia:
@Miciah LGTM.

Since we have no tests for the routing-daemon that would cover this, [merge]

Miciah:
Thanks, openshift-bot!

PR #6327, opened 2015-12-03
[TO_STAGE] - apache lacks permission to access openshift-node-web-proxy log files

dinhxuanvu:
The directory /var/log/openshift/node is currently owned by root under 750 node.
As a result, any other processes will not be able to access that directory which
will cause permission failure.

This commit changes the directory permission to 751 which allows users such as
apache-owned processes to traverse the directory and access to sub-directory
/node-web-proxy which is owned by apache.

Bug 1285424
Link https://bugzilla.redhat.com/show_bug.cgi?id=1285424


dinhxuanvu:
@a13m Would like to have your blessing before merging. Thanks!

PR #6333, opened 2015-12-04
Bug 1082610 - ssh in head gear of scalable app doesn't return any errors

dinhxuanvu:
The HAProxy cartridge in head gear contains a ssh wrapper that is set to
quiet mode (-q flag). As a result, if ssh is not executed correctly,
no errors are shown as they are suppressed bu the -q flag.

This commit will remote the ssh wrapper as it's no longer needed in
HAProxy cartridge. As a result, ssh in head gear will use standard
ssh from /usr/bin/bin which doesn't have -q flag.

Bug 1082610
Link https://bugzilla.redhat.com/show_bug.cgi?id=1082610


dinhxuanvu:
test] [extended:cartridge,node]

dinhxuanvu:
[test] [extended:cartridge,node]

PR #6337, opened 2015-12-08
Bug 1140816 - oo-admin-ctl-district missing documentation for listing districts

dinhxuanvu:
The 'oo-admin-ctl-district' doesn't have option to list all districts. Instead,
if command is executed without any parameters, the command will display a list
of districts. Also, the option 'list-available' is misleading as it displays a
list of available nodes, not districts. The 'oo-admin-ctl-district' experiences
the same behavior with displaying a list of regions when no options are selected.
This issue causes certainly confusions for those two commands.

This commit will explicitly display usage when user runs those commands without
parameters. A 'list' option is added to command list for 'oo-admin-ctl-district'
to allow user to display available districts. Also, 'list-available' is renamed
to nodes-available to clearly state which info will be displayed.

Bug 1140816
Link https://bugzilla.redhat.com/show_bug.cgi?id=1140816

### DIFF
--- a/broker-util/man/oo-admin-ctl-district.txt2man
+++ b/broker-util/man/oo-admin-ctl-district.txt2man
@@ -1,10 +1,10 @@
 NAME
-  oo-admin-ctl-district 
-  
+  oo-admin-ctl-district
+
 SYNOPSIS
-  oo-admin-ctl-district [-h|--help] [-u|--uuid district_uuid] 
-  [-c|--command cmd] [-n|--name district_name] 
-  [-p|--node_profile gear_size] [-i|--server_identity srv_ident] 
+  oo-admin-ctl-district [-h|--help] [-u|--uuid district_uuid]
+  [-c|--command cmd] [-n|--name district_name]
+  [-p|--node_profile gear_size] [-i|--server_identity srv_ident]
   [-s|--size cap_size] [-r|--region region_name] [-z|--zone zone_name]
   [-b|--bypass]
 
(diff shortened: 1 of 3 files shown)


dinhxuanvu:
[test]

dinhxuanvu:
@Miciah @tiwillia Ready to review and merge. Thanks!

dinhxuanvu:
@Miciah @tiwillia Not a high priority PR but  please review and merge if you have the time. Thanks.

tiwillia:
@dinhxuanvu sorry for the late reply here.

This change LGTM, but the oo-admin-ctl-district man page will need to be updated appropriately before we merge this PR:
https://github.com/openshift/origin-server/blob/master/broker-util/man/oo-admin-ctl-district.txt2man

We also need to make sure we modify the relevant documentation. For example, the administration guide mentions using `oo-admin-ctl-district` with no arguments to list the districts. It also includes a copy of the help page, which will need to be updated:

https://github.com/openshift/origin-server/blob/master/documentation/oo_administration_guide.adoc

You'll need to submit a documentation bug for the documentation change.

dinhxuanvu:
@tiwillia manpage file has been updated. I'll fill out a BZ for documentation soon. Thanks!

dinhxuanvu:
@tiwillia Documentation bug is already filed
https://bugzilla.redhat.com/show_bug.cgi?id=1292222

tiwillia:
Lets go ahead and [merge] this then.

PR #6335, opened 2015-12-07
Bug 1160934 - "oo-admin-ctl-gears stopgear" failed to stop idled gear

dinhxuanvu:
During the 'idlegear' operation, the gear in fact is stopped but the
state value is changed to 'idle'. Then, subsequently, when 'stopgear'
operation is excuted, the cartridge doesn't perform any extra works
as the gear is already stopped. However, the state is not modified
to 'stopped' due to lacking of explicit state change in application
container itself.

This commit will add explicit state change to application container
to ensuring the state is changed accordingly.

Bug 1160934
Link https://bugzilla.redhat.com/show_bug.cgi?id=1160934

Signed-off-by: Vu Dinh vdinh@redhat.com

PR #6340, opened 2015-12-16
Bug 1292206 - oo-idler-stats crashes when there is idled gear in the app

dinhxuanvu:
The oo-idler-stats command is currently broken as it crashes with ValueError
if there is idled gear in the application.

This issue is due to PR 6324 on origin-server has removed the unwanted empty
item ('') on the list which is returned by run() method. However, is_idled()
method has not been fixed and still attempts to remove that unwanted item
manually from the list which causes ValueError due to the fact that item is
no longer existed. This commit removes the unnecessary code that attempts
to remove non-exist item from the list in is_idled() method to avoid the
ValueError issue.

Bug 1292206
Link https://bugzilla.redhat.com/show_bug.cgi?id=1292206

### DIFF
--- a/node-util/sbin/oo-idler-stats
+++ b/node-util/sbin/oo-idler-stats
@@ -77,7 +77,6 @@ def is_idled(appid):
         ret, all_idled_apps = run("/usr/sbin/oo-admin-ctl-gears listidle | /bin/cut -d' ' -f1")
         if ret != 0:
             return False  # an error occurred, bail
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
[test] related to #6335 and #6324.
@tiwillia Please review and merge :D. Thanks!

tiwillia:
This LGTM, tested and it appears to resolve the issue.

[merge] assuming tests pass.

PR #6342, opened 2016-01-05
Remove SSL Cert Chain Field from web console

sallyom:
Bug 1268317, Bug 1281901, Bug 1269637
Bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1268317
Bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1281901
Bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1269637

Remove SSL Certificate Chain Field from web console.
Document that the user must concatenate SSL cert files into a single file to upload,
or upload the already-concatenated file included in the SSL certificate from
the SSL certificate provider.

### DIFF
--- a/console/app/models/alias.rb
+++ b/console/app/models/alias.rb
@@ -17,41 +17,29 @@ class Alias < RestApi::Base
   custom_id :id
 
   # Optional accessors for form upload
-  attr_accessor :certificate_file, :certificate_chain_file, :certificate_private_key_file
+  attr_accessor :certificate_file, :certificate_private_key_file
 
   belongs_to :application
 
   alias_attribute :name, :id
   alias_attribute :certificate, :ssl_certificate
-  alias_attribute :certificate_chain, :ssl_certificate_chain
   alias_attribute :certificate_private_key, :private_key
   alias_attribute :certificate_pass_phrase, :pass_phrase
 
-  validates_presence_of :certificate_file, :message => "Required if a certificate chain or private key file is selected.", :if => lambda { |a| a.certificate_chain_file.present? || a.certificate_private_key_file.present? }
+  validates_presence_of :certificate_file, :message => "Required if a private key file is selected.", :if => lambda { |a| a.certificate_private_key_file.present? }
   validates :certificate_file, :length => {:minimum => 1, :message => "SSL certificate file was empty."}, :allow_nil => true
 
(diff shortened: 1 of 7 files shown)


sallyom:
[test]

sallyom:
[test]

sallyom:
@fabianofranz @jwforres when you have a chance -ha!- can you review this?  Any problem with a single field for SSL cert upload (no cert chain file field) w/ documentation explaining that users need to concatenate cert files?  See referenced 3 bugs above - concatenating certs is the workaround for these, but do you know of any cases where this won't work/cert files must be kept separate?  Thanks!

fabianofranz:
@sallyom there are no cases where files must be kept separate (they are always sent concatenated to the node), so I'm fine with this. The help block must be very clear about what users have to do, maybe we should add a link to some FAQ or documentation that explains how to do the concatenation properly? Examples [here](https://www.digicert.com/ssl-support/pem-ssl-creation.htm) and [here](https://support.comodo.com/index.php?/Default/Knowledgebase/Article/View/643/17/).

fabianofranz:
@sallyom also note that this will require some docs updates, for example https://developers.openshift.com/en/managing-domains-ssl.html#_web_console_2.

sallyom:
Awesome, thanks @fabianofranz!  @tiwillia @Miciah, I'll look at the docs/help block, sound good to you?

tiwillia:
@sallyom Sounds good.

The help-block looks good so far, although I would echo Fabiano's concern that we should probably include a link to the relevant documentation or one of the FAQs on how to concatenate these files.

sallyom:
@tiwillia @Miciah @fabianofranz I think I've addressed all feedback.  I've submitted doc changes (https://github.com/openshift/devcenter/pull/406), the link in the help block won't be helpful unless this PR is also merged.

sallyom:
@tiwillia @Miciah I think this PR is good to go, I've looked at documentation and I can't see any need for  doc changes/additions regarding the SSL cert chain.  I think the help block I added is/was the only 'documentation' regarding how to upload SSL certs.  Thanks for your review.

tiwillia:
LGTM @sallyom 

Sorry for the late response on this. Lets [merge]

tiwillia:
Lets try that one more time to make sure these are spurious failures. They appear to be unrelated at least.

re-[merge]

tiwillia:
Unrelated test failures have been resolved, lets try that again! 

re-re-[merge]

PR #6343, opened 2016-01-05
oo-admin-ctl-team: small fixes

thrasher-redhat:
check to avoid output if no members

Bug 1123337
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1123337

The oo-admin-ctl-team script was printing "MEMBER|ADD|groupname|" even if there
were no members being added.  This will now only output if there are members in
the members_to_add variable.

Also included is a commit to fix a variable typo which should mean a more accurate select statement.


Miciah:
Looks good.  Thanks!

PR #6345, opened 2016-01-08
oo-admin-cartridge: remove duplicate results

thrasher-redhat:
Bug 1111105
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1111105

Fixes double output for the oo-admin-cartridge --action list --debug command.
Output will now avoid printing a second copy of the cartridge without a
cartridge version number.

The way the code is written, there will sometimes be an extra '>' printed - however the entire cartridge object will no longer be printed twice.


thrasher-redhat:
[test]

tiwillia:
Tested and this LGTM. The fix does appear to resolve the issue.

Lets [merge]!

PR #6346, opened 2016-01-11
Fix app info search by app name bug

joelsmith:
The bug causes only the final matching application to be shown
when searching by app name.


Miciah:
Looks fine.  Thanks!

openshift-bot, please [merge].

Miciah:
Thanks, openshift-bot!

PR #6347, opened 2016-01-13
oo-accept-node: update to check cgroup tasks

thrasher-redhat:
Bug 1067107
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1067107

Updates oo-accept-node to check thread ids against cgroup tasks instead of
checking process ids against cgroup procs.

### DIFF
--- a/node-util/sbin/oo-accept-node
+++ b/node-util/sbin/oo-accept-node
@@ -436,12 +436,12 @@ end
 
 
 #
-# Check gear processes belong to cgroups
+# Check gear threads belong to cgroups
 #
-def check_cgroup_procs
-    verbose("checking cgroups processes")
+def check_cgroup_tasks
+    verbose("checking cgroups tasks")
 
-    ### Gather current procs running ###
+    ### Gather current threads running ###
     min_uid = $CONF.get('GEAR_MIN_UID').to_i
     max_uid = $CONF.get('GEAR_MAX_UID').to_i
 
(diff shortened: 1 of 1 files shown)


thrasher-redhat:
[test]

node-util/sbin/oo-accept-node
```diff
@@ -455,10 +455,11 @@ def check_cgroup_procs
 
     verbose("determining node uid range: #{min_uid} to #{max_uid}")
 
-    all_user_procs = %x[/bin/ps -e -o uid,pid].split("\n")
-    ps_procs = Hash.new{|hash, key| hash[key] = Array.new}
-    all_user_procs.each do |line|
-        uid,pid = line.split[0,2]
+    # Bug 1067107 - Check tids against tasks instead of pids against cgroup procs
+    all_user_threads = %x[/bin/ps -e H -o uid,pid,tid].split("\n")
+    ps_threads = Hash.new{|hash, key| hash[key] = Hash.new{|hash, key| hash[key] = ""}}
```
Miciah: I don't see a reason for the inner Hash's initializer, `{|hash, key| hash[key] = ""}`.  If there is some reason you need it, maybe you could rename the parameters `hash` and `key` to, say, `h` and `k` to avoid shadowing the outer initializer's parameters.

thrasher-redhat:
@a13m Wanted to run this by you before merging to make sure that this is what you had in mind.  Don't want to introduce a major change into a commonly used script like oo-accept-node without lots of input.

Miciah:
Looks spiffin'.  Please [merge]!

Miciah:
Yay! Thanks!

PR #6348, opened 2016-01-14
admin_helper.rb: check for nil plans

thrasher-redhat:
Bug 1279584
BZ https://bugzilla.redhat.com/show_bug.cgi?id=1279584

In situations where get_plans() returns nil when trying to determine user
capabilities, attempting to populate_user_hash() in oo-admin-chk and
oo-admin-repair would result in the script crashing due to an undefined method
exception.  The script will now check if get_plans() comes back nil and will drop
back to using the default user capabilities, as well as displaying a warning
that the plan has not been implemented.  This warning will not cause
oo-admin-chk or oo-admin-repair to fail.

### DIFF
--- a/controller/app/helpers/admin_helper.rb
+++ b/controller/app/helpers/admin_helper.rb
@@ -25,6 +25,7 @@ module AdminHelper
   $summary_count = 0
   $total_errors = 0
   $verbose = false
+  $invalid_plan = false
 
   $billing_enabled = Rails.configuration.respond_to?('billing')
   $districts_enabled = Rails.configuration.msg_broker[:districts][:enabled]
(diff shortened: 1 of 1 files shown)


controller/app/helpers/admin_helper.rb
```diff
@@ -99,8 +99,14 @@ def get_user_info(user)
     user["ssh_keys"].each { |k| user_ssh_keys << ["#{user['_id'].to_s}-#{k['name']}", Digest::MD5.hexdigest(k["content"]), nil] if k["content"] } if user["ssh_keys"].present?
     user_caps = user["capabilities"]
     if user["plan_id"]
-      plan_caps = OpenShift::BillingService.instance.get_plans[user["plan_id"].to_sym][:capabilities]
-      user_caps = plan_caps.deep_dup.merge(user_caps)
+      plans = OpenShift::BillingService.instance.get_plans[user["plan_id"].to_sym]
```
tiwillia: I believe this will still cause the error to be reported when `get_plans` returns `nil`. You cannot call `[]` on `nil`. Instead, call `[user["plan_id"].to_sym]` after you ensure that `get_plans` did not return `nil`.

tiwillia: Additionally, the use of `plans` as the variable name is confusing, since this would only return one plan.

controller/app/helpers/admin_helper.rb
```diff
@@ -99,8 +99,14 @@ def get_user_info(user)
...
+      if plans.present?
+        plan_caps = plans[:capabilities]
+        user_caps = plan_caps.deep_dup.merge(user_caps)
+      else
+        puts "Warning: The plan_id: #{user["plan_id"]} exists, but does not have a valid plan.  Default user capabilities will be used."
+        puts "Has get_plans in the billing provider been implemented?"
```
tiwillia: It looks like this warning is going to be displayed for every user. On a system with >5k users where all users have a plan_id but `get_plans` is not implemented, that is a lot of unnecessary output.

Maybe move the warning out of the loop and state something like the following:
```
Warning: At least one user has a plan id defined but no matching plans could be found. User capabilities defined by the plan will not be taken into consideration.
```

controller/app/helpers/admin_helper.rb
thrasher-redhat: I can definitely change plans to plan to make the variable more accurate.  I'm pretty sure that this is correct though for the call to get_plans().  The user["plan_id"] is going to come back with a string such as "free".  It converts that to a symbol :free, and then runs checks get_plans[:free] and sets that to plan, which we're looking at to see if it is nil.

tiwillia:
LGTM

tiwillia:
Although there are no tests for oo-admin-chk, it may still be useful to run through the testing suite since this is a change to the controller code.

[test]

tiwillia:
[merge] please!

PR #6350, opened 2016-01-20
oo-accept-node: check_quotas: cope with loop mount

Miciah:
#### `oo-accept-node`: Add and use `do_warn`

Define and use the `do_warn` method.
#### `oo-accept-node`: `check_quotas`: cope with loop mount

Modify `oo-accept-node`'s quota check to check whether the mount is using a loop device mount and, if so, check the quota on the loop device.  In addition, print a warning that using a loop mount may impact performance negatively.

This commit fixes bug 1265811.

https://bugzilla.redhat.com/show_bug.cgi?id=1265811
#### `oo-accept-node`: `check-quotas`: simplify code

Simplify the logic in `oo-accept-node`'s `check_quotas` function to run the `df` command and parse its output only once.
## 

@sferich888, @tiwillia

I could not resist making the gratuitous cleanup, but I can drop it if you think I ought.

### DIFF
--- a/node-util/sbin/oo-accept-node
+++ b/node-util/sbin/oo-accept-node
@@ -547,6 +547,12 @@ def check_quotas
     verbose("checking filesystem quotas")
     oo_device=%x[/bin/df -P #{$GEAR_BASE_DIR} | /usr/bin/tail -1].split[0]
     oo_mount=%x[/bin/df -P #{$GEAR_BASE_DIR} | /usr/bin/tail -1 | /usr/bin/tr -s ' '].split[5]
+    awk = %[{ if ($1 == "#{oo_device}" && $3 == "#{oo_mount}") { print $6 } }]
+    fs_options = %x[mount | /bin/awk '#{awk}'].split('\n').last rescue ""
+    if fs_options.match /loop=([^,]+)/
+      oo_device = $1
+      do_warn "#{oo_mount} is a loop mount (loop device: #{oo_device}).  Using a loop mount may reduce performance."
+    end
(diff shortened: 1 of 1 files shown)


Miciah:
Please [test]!

tiwillia:
@Miciah do we need both commits here, or can we squash them into one?

This LGTM otherwise.

Miciah:
I'd rather keep them in separate commits, or even drop the simplification, rather than squash, which I believe would make the changes more difficult to read.

sferich888:
Is it really wise to support loop devices for the /var/lib/openshift directory. Wouldn't this cause performance issues if allowed?

Miciah:
There probably generally is a better solution to any problem that someone solves using loop devices.

If an administrator does use a loop device for `/var/lib/openshift/`, `oo-accept-node` currently prints a bogus failure.  This PR enables `oo-accept-node` to function properly.  An alternative would be to make `oo-accept-node` clearly and explicitly fail (or print a warning) if `/var/lib/openshift/` is on a loop device, with an accurate error message.  Would that alternative be better?

sferich888:
I propos allowing for this and posting a clear warning

Miciah:
I have updated the test to print a warning if `/var/lib/openshift/` is a loop mount:

```ruby
      do_warn "#{oo_mount} is a loop mount (loop device: #{oo_device}).  Using a loop mount may reduce performance." 
```

@tiwillia, does this look good to you?

tiwillia:
@Miciah :+1: LGTM

Shall I merge?

Miciah:
Thanks! I do the honours...

PR #6351, opened 2016-01-22
Bug 1296943: Quickstart deployment via git fails to bind address

dinhxuanvu:
When quickstart is deployed with rhc --from-code, there is a start process
(from control script) called during post_receive step which starts the gear and
apache processes. However, in the next step, the cartridge is started as well.
As a result, there are two consecutive start processes leading to the second one
fails due to the fact apache (httpd) process is already started and currently
in use.

This commit adds a check of current state of the gear before starting the cartridge.
If the gear is already started, the restart cartridge process is triggered instead.
Otherwise, the start process is striggered as expected. This fix only change the
behavior initial build of cartridge during git deployment and it's operating within
the local gear. Therefore, it doesn't impact any other gears that are running in
the same node.

Bug 1296943
Link https://bugzilla.redhat.com/show_bug.cgi?id=1296943

### DIFF
--- a/node/lib/openshift-origin-node/model/v2_cart_model.rb
+++ b/node/lib/openshift-origin-node/model/v2_cart_model.rb
@@ -400,7 +400,11 @@ module OpenShift
           if cartridge.name == primary_cartridge.name and empty_repository?
             output << "CLIENT_MESSAGE: An empty Git repository has been created for your application.  Use 'git push' to add your code."
           else
-            output << start_cartridge('start', cartridge, user_initiated: true)
+            if (@state.value == State::STARTED)
+              output << start_cartridge('restart', cartridge, user_initiated: true)
+            else
+              output << start_cartridge('start', cartridge, user_initiated: true)
+            end
           end
           output << cartridge_action(cartridge, 'post_install', software_version)
         end
--- a/documentation/oo_cartridge_developers_guide.adoc
+++ b/documentation/oo_cartridge_developers_guide.adoc
@@ -74,13 +74,13 @@ An example `manifest.yml` file:
 Name: PHP
 Cartridge-Short-Name: PHP
 Cartridge-Version: '1.0.1'
-Compatible-Versions: 
+Compatible-Versions:
   - '1.0.1'
 Cartridge-Vendor: redhat
 Display-Name: PHP 5.3
 Description: "PHP is a general-purpose server-side scripting language..."
 Version: '5.3'
-Versions: 
+Versions:
   - '5.3'
 License: "The PHP License, version 3.0"
 License-Url: http://www.php.net/license/3_0.txt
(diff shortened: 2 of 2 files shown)


dinhxuanvu:
@tiwillia @jwhonce [test] and ready for review. This is the least invasive fix that I can think of at the moment to avoid regression issues or unexpected consequences down the road. Thanks!

node/lib/openshift-origin-node/model/v2_cart_model.rb
```diff
@@ -400,7 +400,11 @@ def post_configure(cartridge_name)
           if cartridge.name == primary_cartridge.name and empty_repository?
             output << "CLIENT_MESSAGE: An empty Git repository has been created for your application.  Use 'git push' to add your code."
           else
-            output << start_cartridge('start', cartridge, user_initiated: true)
+            if @state.value == State::STARTED
+              output << start_cartridge('restart', cartridge, user_initiated: true)
```
jwhonce: FYI -- restart may not be equivalent to stop/start. Therefore you are introducing a different semantic to post_configure using restart. The issue with calling stop/start is since post_install has not run yet, stop may fail with an error.

dinhxuanvu:
re-[test] @jwhonce As we discussed last Friday, I have changed the code to explicitly stop the cartridge instead of restarting it.

node/lib/openshift-origin-node/model/v2_cart_model.rb
```diff
@@ -400,6 +400,9 @@ def post_configure(cartridge_name)
           if cartridge.name == primary_cartridge.name and empty_repository?
             output << "CLIENT_MESSAGE: An empty Git repository has been created for your application.  Use 'git push' to add your code."
           else
+            if @state.value == State::STARTED
+              output << stop_cartridge(cartridge, user_initiated: true)
+            end
             output << start_cartridge('start', cartridge, user_initiated: true)
```
tiwillia: Why are we stopping the cartridge above? Why not just throw the inverse of that if statement around this line, so that the cartridge stays started and doesn't get re-started? Seems silly to start-stop-start when we could just start once.

dinhxuanvu: @tiwillia  I see your point. Looked at the code and it seems like during post-receive step, the start gear process does use start cartridge as well. So I dont think we miss anything if we skip the second start process.

dinhxuanvu:
@jwhonce A slight change of code :), as Tim suggested to skip the start cartridge entirely which I thought so too but I didn't do so as I wasn't sure if it has any consequences for that. Turns out after looking at the code again, the start gear in post-receive does call start cartridge as well. So, it doesn't miss anything if we skip the second start cartridge if the cartridge is already started according to the state value.

dinhxuanvu:
re-[test]

dinhxuanvu:
re-[test]

dinhxuanvu:
re-[test]

dinhxuanvu:
@jwhonce Would like to have your review on this PR before merging :). I know you are a not big fan of restart approach :D but so far this is only solution that is not causing issues with Jenkins test + addon cartridges. I have tried to stop+start and skipping start entirely and both causes failures on Jenkins and on my manual testing as well, especially when adding secondary cartridge (such as mysql) to gear. The restart approach, on the other hand, is working well for Jenkins + my manual test as it uses restart method in control script which is to me at least is safer as it's written for that purposes.

jwhonce:
@dinhxuanvu  If we're going with restart then I believe you should add something to the oo_user_guide.adoc in the action_hooks section about this side case.  At least then we can point to the documentation.  Thanks

dinhxuanvu:
@tiwillia Ready for review/merge. Thanks :)

tiwillia:
LGTM [merge]

PR #6353, opened 2016-01-27
oo-init-quota: allow tab delimiters in /etc/fstab

### LINKED ISSUE #6352: oo-init-quota does not set mount options for tab-delimited /etc/fstab
If the entry containing the mountpoint for /var/lib/openshift in /etc/fstab is tab-delimited, `oo-init-quota` will not add the quota options `usrjquota=aquota.user,jqfmt=vfsv0` to the entry.

charlessimpson:
Allow tab delimiters in /etc/fstab when adding quota options to the OpenShift mountpoint.

Fixes #6352.

### DIFF
--- a/node-util/sbin/oo-init-quota
+++ b/node-util/sbin/oo-init-quota
@@ -66,7 +66,7 @@ function get_fstab_mount_options() {
 # get the current mount options for a given mount point
 function get_current_mount_options() {
     # options are the comma delimited string between parens
-    mount | egrep " $1 " | sed -e 's/^.*(// ; s/).*$// ' | sort -u
+    mount | egrep "\s+$1\s+" | sed -e 's/^.*(// ; s/).*$// ' | sort -u
 }
 
 ORIGIN_MOUNTPOINT=$(get_mountpoint $origin_dir)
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
[test]

node-util/sbin/oo-init-quota
```diff
@@ -79,11 +79,11 @@ QUOTA_OPTIONS=usrjquota=aquota.user,jqfmt=vfsv0
 function update_fstab() {
     # ORIGIN_MOUNTPOINT=$1
 
-    if ! tr -s ' ' < /etc/fstab | egrep "^[^#].+ $1 ext4 [^ ]+$QUOTA_OPTIONS "; then
```
Miciah: Interesting: `[^ ]+` causes this old regex to fail to match if the quota options are not preceded by other options, so this change avoids adding duplicate `$QUOTA_OPTIONS` in that case.

node-util/sbin/oo-init-quota
```diff
@@ -66,7 +66,7 @@ function get_fstab_mount_options() {
 # get the current mount options for a given mount point
 function get_current_mount_options() {
     # options are the comma delimited string between parens
-    mount | egrep " $1 " | sed -e 's/^.*(// ; s/).*$// ' | sort -u
+    mount | egrep "\s+$1\s+" | sed -e 's/^.*(// ; s/).*$// ' | sort -u
```
Miciah: `mount` shouldn't output tabs, so this change shouldn't be necessary, but then again, it shouldn't hurt anything.

node-util/sbin/oo-init-quota
```diff
@@ -79,11 +79,11 @@ QUOTA_OPTIONS=usrjquota=aquota.user,jqfmt=vfsv0
 function update_fstab() {
     # ORIGIN_MOUNTPOINT=$1
 
-    if ! tr -s ' ' < /etc/fstab | egrep "^[^#].+ $1 ext4 [^ ]+$QUOTA_OPTIONS "; then
+    if ! tr -s '[:blank:]' < /etc/fstab | egrep "^[^#].+\s+$1\s+ext4\s+.*$QUOTA_OPTIONS\s+"; then
```
Miciah: Could we change the last `\s+` to `[[:blank:],]+`? Otherwise, this pattern fails to match if `$QUOTA_OPTIONS` is followed by further options.

tiwillia: The `.*` before `$QUOTA_OPTIONS` will pick up any character, even spaces. This should probably be `[^ ]*` or even better `[^\h]` (which is shorthand for `[^[:blank:]]`). The `[^ ]+` previously in this line is odd and probably incorrect. It requires a non-space character before the quota options when it is possible that the quota options may be first in the list of options.

tiwillia:
Lets re-[test], as the test failures are unrelated.

PR #6355, opened 2016-01-27
Bug 1292445: Out of memory error due to insufficient mysql sort buffer size

dinhxuanvu:
The new SilverStripe CMS app using php and mysql cartridges fails due to out
of memory on mysql sort buffer which is currently set at 128K.

This commit increases the sort buffer size to 256K and also add the environment
variable OPENSHIFT_MYSQL_SORT_BUFFER_SIZE to allow user to modify the 
mysql sort buffer size at their own choices without having modify the my.cnf directly.

Bug 1292445
Link https://bugzilla.redhat.com/show_bug.cgi?id=1292445

### DIFF
--- a/cartridges/openshift-origin-cartridge-mysql/conf/my.cnf.erb
+++ b/cartridges/openshift-origin-cartridge-mysql/conf/my.cnf.erb
@@ -8,6 +8,9 @@
 # max_connections -> OPENSHIFT_MYSQL_MAX_CONNECTIONS
 # ft_min_word_len -> OPENSHIFT_MYSQL_FT_MIN_WORD_LEN
 # ft_max_word_len -> OPENSHIFT_MYSQL_FT_MAX_WORD_LEN
+# max_allowed_packet -> OPENSHIFT_MYSQL_MAX_ALLOWED_PACKET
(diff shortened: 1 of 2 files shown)


dinhxuanvu:
[test]

tiwillia:
This looks good, do we need to document the new environment variable somewhere?

dinhxuanvu:
It has been indicated in the my.cnf file itself so no need to change document anywhere else just like the rest of environment variables on that file :)

dinhxuanvu:
Ah, never mind, look like the comment section needs to get updated a bit. Will take care of that.

dinhxuanvu:
re-[test]

tiwillia:
@dinhxuanvu that is probably good enough. But we do have documentation on specific environment variables available for different cartridges:

https://docs.openshift.org/origin-m4/oo_cartridge_guide.html#environment-variables-7

Looks like it isn't a complete list. Perhaps needs some updating in another PR.

dinhxuanvu:
Fixed and re-[test]

tiwillia:
@dinhxuanvu the changes look great, thanks! [merge]

dinhxuanvu:
re-[test]

PR #6358, opened 2016-02-02
Bug 1104750 - oo-admin-chk --level 0 takes too long in prod

dinhxuanvu:
The current threshold for gear creation time is 10 minutes (600 seconds)
which ends up generating a lot of false errors during oo-admin-chk run
as the command takes a long time to complete and during that time period
gears can be deleted and created.

This commit increases the threshold to 6 hours (21600 seconds) to reduce
the false errors generated oo-admin-chk.

Bug 1104750
Linkhttps://bugzilla.redhat.com/show_bug.cgi?id=1104750

### DIFF
--- a/broker-util/oo-admin-chk
+++ b/broker-util/oo-admin-chk
@@ -138,7 +138,7 @@ $datastore_hash.each do |gear_uuid, gear_info|
   server_identity = gear_info['server_identity']
   app_id = gear_info['app_id']
 
-  if (current_time - creation_time) > 600
+  if (current_time - creation_time) > 21600
     if !node_hash.has_key? gear_uuid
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@a13m [test]. We increase the threshold to 6 hours and that would reduce the false positives on the error report from oo-admin-chk. Unless you have a different number in mind, would like to get your :+1:  before merging. Thanks :)

dinhxuanvu:
@tiwillia @Miciah Have checked with agrimm and confirmed the new time threshold is good. Please bless this PR with a merge tag please :)

tiwillia:
Excellent, thanks for this @dinhxuanvu. Lets go ahead and [merge] since we have passing tests and the :+1: from @a13m

PR #6360, opened 2016-02-04
Bug 1299095 - oo-diagnostic test_broker_certificate fails on broker

dinhxuanvu:
The 'test_broker_certificate' method in oo-diagnostics fails if it's run
on broker due to the fact the test will checking for a file that only
exists on node but not broker.

This is a regression issue as in PR 6275, several lines of code was added
which was intended to run on the node only. As a result, when the code was
run in the broker, the failure occurs. This commit will separate the code
for broker and node test into 2 separate tests: test_broker_certificate
and test_node_certificate.

Bug 1299095
Link https://bugzilla.redhat.com/show_bug.cgi?id=1299095

### DIFF
--- a/common/bin/oo-diagnostics
+++ b/common/bin/oo-diagnostics
@@ -1216,7 +1216,7 @@ class OODiag
     else
       mode = "test"
     end
-    
+
     msg_log = `grep 'Broker key authentication failed' #{log_dir + mode}.log`
 
     unless msg_log.empty?
@@ -1236,6 +1236,7 @@ class OODiag
     end
   end
 
+  # Test certificate from the broker
   def test_broker_certificate
     skip_test unless @is_broker
 
@@ -1257,30 +1258,6 @@ class OODiag
     if subject == issuer
       do_warn "Using a self-signed certificate for the broker"
     end
-
-    broker_host_env = File.read("/etc/openshift/env/OPENSHIFT_BROKER_HOST")
-
-    # Retrieve the SSL cert from broker_host_env
-    env_response = `curl -k -s -v https://#{broker_host_env.chomp}/ -o /dev/null 2>&1`
-    if $? != 0
-      do_fail <<-CANNOTCONNECT
-        Attempt to connect to #{broker_host_env} but failed. curl returned with
-        #{env_response.strip.empty? ? "no output." : "the following output:\n" + env_response}
-
-        Please check whether the httpd service in #{broker_host_env} is running.
-      CANNOTCONNECT
-      return
-    end
-
-    # Get SSL cert's issuer, subject and commonname
-    env_issuer = env_response.slice(/^\*\s*issuer: .*$/).gsub(/^\*\s*issuer: /,"")
-    env_subject = env_response.slice(/^\*\s*subject: .*$/).gsub(/^\*\s*subject: /,"")
-    env_commonname = env_response.slice(/^\*\s*common name: .*$/).gsub(/^\*\s*common name: /,"")
-
-    # Check to see if SSL certs from current broker and BROKER_HOST match
-    do_fail "SSL cert issuer doesn't match between localhost and #{broker_host_env} defined in /etc/openshift/env/OPENSHIFT_BROKER_HOST" if env_issuer != issuer
-    do_fail "SSL cert subject doesn't match between localhost and #{broker_host_env} defined in /etc/openshift/env/OPENSHIFT_BROKER_HOST" if env_subject != subject
-    do_fail "SSL cert commonname doesn't match between localhost and #{broker_host_env} defined in /etc/openshift/env/OPENSHIFT_BROKER_HOST" if env_commonname != commonname
 
     apacheconfig = `httpd -S 2> /dev/null`.slice(  / ^\*:443.*  (\n^\s.*)*  \n(\S|\z) /x  )
     servername = apacheconfig.scan(/(?:(?:default server )|(?:port 443 namevhost ))(\S+) \((?:[^:]+)/)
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@tiwillia @Miciah [test] This is an regression issue as the test is intended to run on the node, not the broker. Would like to expedite review/merge process to make it to 2.2.9 release. Thanks.

dinhxuanvu:
Hi all. Please put a hold on this. I need to do some further checking on this issue. Sorry about that.

tiwillia:
@dinhxuanvu FWIW, I think you may be correct. `/etc/openshift/env` is a node directory, and the test does look to be curling the reverse proxy at `localhost`, which is only located on a node. Additionally, its curling the broker using the address in the environment variable, rather than localhost.

dinhxuanvu:
@tiwillia Well, I was just checking some stuff and the change was originally added by me but the @is_broker was not mine and I think I didn't change it. It's OK by you and we can merge after the test is completed.

tiwillia:
@dinhxuanvu I am fairly confident this fix is proper. After doing some testing on a cluster where the broker is not co-located with the node, I can confirm that this test was meant for the node. 

Lets run this by ops real fast to make sure it isn't going to throw any false positives though, since we may not know whether the test properly tests the broker certificate.

dinhxuanvu:
@a13m Would like to run this by you :)...

Miciah:
The test is curling localhost to get the broker certificate, but localhost will be the node if the test is run on a node.

tiwillia:
Since ops doesn't use this and testing appears to result in proper results being shown on the broker, lets go ahead and [merge] this.

dinhxuanvu:
@tiwillia @Miciah Ready for review again. Thanks.

common/bin/oo-diagnostics
```diff
@@ -1307,6 +1284,49 @@ class OODiag
...
+      do_warn "Using a self-signed certificate for the node"
+    end
+
+    broker_host_env = File.read("/etc/openshift/env/OPENSHIFT_BROKER_HOST")
+
+    # Retrieve the SSL cert from broker_host_env
+    env_response = `curl -k -s -v https://#{broker_host_env.chomp}/ -o /dev/null 2>&1`
+    if $? != 0
+      do_fail <<-CANNOTCONNECT
+        Attempt to connect to #{broker_host_env} but failed. curl returned with
```
Miciah: s/ but// or s/Attempt/Attempted/

common/bin/oo-diagnostics
```diff
@@ -1307,6 +1284,49 @@ class OODiag
...
+        #{env_response.strip.empty? ? "no output." : "the following output:\n" + env_response}
+
+        Please check whether the httpd service in #{broker_host_env} is running.
+      CANNOTCONNECT
+      return
+    end
+
+    # Get SSL cert's issuer, subject and commonname
```
Miciah: This comment is no longer accurate.

common/bin/oo-diagnostics
```diff
@@ -1307,6 +1284,49 @@ class OODiag
...
+    env_issuer = env_response.slice(/^\*\s*issuer: .*$/).gsub(/^\*\s*issuer: /,"")
+
+    # Check to see if SSL certs from current broker and BROKER_HOST match
```
Miciah: This comment is no longer accurate.

common/bin/oo-diagnostics
```diff
@@ -1307,6 +1284,49 @@ class OODiag
...
+    issuer = response.slice(/^\*\s*issuer: .*$/).gsub(/^\*\s*issuer: /,"")
+    subject = response.slice(/^\*\s*subject: .*$/).gsub(/^\*\s*subject: /,"")
+    if subject == issuer
+      do_warn "Using a self-signed certificate for the node"
+    end
+
+    broker_host_env = File.read("/etc/openshift/env/OPENSHIFT_BROKER_HOST")
+
+    # Retrieve the SSL cert from broker_host_env
+    env_response = `curl -k -s -v https://#{broker_host_env.chomp}/ -o /dev/null 2>&1`
```
Miciah: Maybe rename `env_*` to `broker_*` now that we are comparing the broker host with the node host (whereas when you wrote the comment, we were comparing the broker host that was specified with the `env/` file with the local host, which we assumed was the same broker host).

common/bin/oo-diagnostics
```diff
@@ -1307,6 +1284,49 @@ class OODiag
...
+    if $? != 0
+      do_fail <<-CANNOTCONNECT
+        Attempt to connect to #{broker_host_env} but failed. curl returned with
+        #{env_response.strip.empty? ? "no output." : "the following output:\n" + env_response}
+
+        Please check whether the httpd service in #{broker_host_env} is running.
```
Miciah: s/in/on/

common/bin/oo-diagnostics
```diff
@@ -1307,6 +1284,49 @@ class OODiag
...
+
+        Please check whether the httpd service on #{broker_host_env} is running.
+      CANNOTCONNECT
+      return
+    end
+
+    # Get SSL cert's issuer
+    broker_issuer = broker_response.slice(/^\*\s*issuer: .*$/).gsub(/^\*\s*issuer: /,"")
+
+    # Check to see if SSL cert issuer from current broker and BROKER_HOST match
```
Miciah: s/current broker/node/

We're comparing `broker_issuer` with `issuer`, which came from `curl`ing localhost, which is the node, right?

tiwillia:
I tested this on a cluster where all components are separated. I can confirm that each of these tests are only run on the appropriate systems. The test functionality for the broker is unchanged and the test functionality for the node appears to work for what I tested.

Lets go ahead and [merge] this again.

Thanks for fixing this @dinhxuanvu and sorry for the confusion.

Miciah:
We're seeing these failures from the acceptance script:

```
FAIL: directory 56b92869ebf9622dcf00007a/.env doesn't have OPENSHIFT_PRIMARY_CARTRIDGE_DIR
```

These failures do not appear to be related to the change in this PR, so let's try to [merge] it again!

dinhxuanvu:
@tiwillia The second merge from @Miciah also failed. This may be related to the stopped Jenkins job the other day.

dinhxuanvu:
re-[test]

PR #6363, opened 2016-02-12
Gears exceeding quota should be allowed to stop, start, move

sallyom:
bz 1292133
https://bugzilla.redhat.com/show_bug.cgi?id=1292133

Create a 1% quota buffer so that gears exceeding allowed quota can still be moved,
stopped or started.

### DIFF
--- a/node/test/node_functional/node_func_test.rb
+++ b/node/test/node_functional/node_func_test.rb
@@ -26,6 +26,9 @@ module OpenShift
         super
 
         @uid = 5994
+        # quota buffer multiplier is configured on broker
+        @quota_blocks_buffer = 1.2
+        @quota_inodes_buffer = 1.2
 
         @config.stubs(:get).with("GEAR_BASE_DIR").returns(GEAR_BASE_DIR)
         @config.stubs(:get).with("GEAR_GECOS").returns('Functional Test')
@@ -69,17 +72,20 @@ module OpenShift
       end
 
       def test_set_quota_fail_quota
+        # test won't fail until blocks exceed blocksmax * blocks buffer
         results = Runtime::Node.get_quota(@uuid)
+
         assert_raises(NodeCommandException) do
-          OpenShift::Runtime::Node.set_quota(@uuid, results[:blocks_used] - 1, '')
+          OpenShift::Runtime::Node.set_quota(@uuid, (results[:blocks_used] - 1) * @quota_blocks_buffer, '')
         end
       end
 
       def test_set_quota_fail_inodes
+        # test won't fail until inodes exceed inodesmax * inodes buffer
         results = Runtime::Node.get_quota(@uuid)
 
         assert_raises(NodeCommandException) do
-          OpenShift::Runtime::Node.set_quota(@uuid, results[:blocks_used], results[:inodes_used] - 1)
+          OpenShift::Runtime::Node.set_quota(@uuid, results[:blocks_used], (results[:inodes_used] - 1) * @quota_inodes_buffer)
         end
       end
 
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -34,7 +34,11 @@ module OpenShift
     # * id: string - a unique app identifier
     # * district: <type> - a classifier for app placement
     #
-    def initialize(id, district=nil)
+    def initialize(id, district=nil, blocks_multiplier=1, inodes_multiplier=1)
+      blocks_multiplier=Rails.configuration.msg_broker[:quota_blocks_buffer]
+      inodes_multiplier=Rails.configuration.msg_broker[:quota_inodes_buffer]
+      @blocks_multiplier = blocks_multiplier
+      @inodes_multiplier = inodes_multiplier
       @id = id
       @district = district
       @disable_print_debug = false
@@ -1768,6 +1772,46 @@ module OpenShift
     end
 
     def update_cluster(gear, options)
+      app = gear.application
+      head_gear_quota = get_quota(app)
+      # set head_gear quotas if necessary
+      # will be necessary if trying to move a non-primary gear before bumping
+      # an over-the-quota-limit head gear
+      #
+      # Find default quotas for node profile
+      default_container = ApplicationContainerProxy.find_available(options)
+      default_quota = default_container.rpc_get_facts_direct(["quota_blocks", "quota_files"])
+      default_blocks = Integer(default_quota[:quota_blocks])
+      default_inodes = Integer(default_quota[:quota_files])
+
+      head_blocks_current_limit = Integer(head_gear_quota[3])
+      head_blocks_used = Integer(head_gear_quota[1])
+
+      head_inodes_current_limit = Integer(head_gear_quota[6])
+      head_inodes_used = Integer(head_gear_quota[4])
+
+      # if head_gear quota usage > 98% current limit, then update-cluster wil need a
+      # small buffer to complete.
+      if head_blocks_used > head_blocks_current_limit * 0.98 || head_inodes_used > head_inodes_current_limit
+        # Use a multiplier of 1.01 to avoid moves failing due to lack of disk space for rewrite of haproxy.cfg and add 10 inodes to inode limit (will be reset to default if usage < default limits).
+        head_blocks_increment = head_blocks_current_limit * 1.01
+        if head_blocks_increment > default_blocks * @blocks_multiplier.to_f
+          head_blocks_increment = default_blocks * @blocks_multiplier.to_f
+        end
+        set_quota(app, head_blocks_increment/1024/1024, head_inodes_current_limit + 10)
+        log_debug "WARNING: Head gear exceeds quota limit, using quota buffer for update-cluster."
+        log_debug "DEBUG: New gear blocks limit:#{head_blocks_increment.round}, new inodes limit: #{head_inodes_current_limit + 10}"
+      else
+        # reset quotas to default if previously bumped but are now under default limits
+        # or if usage < default limit
+        if head_blocks_used < default_blocks
+          head_blocks_current_limit = default_blocks
+        end
+        if head_inodes_used < default_inodes
+          head_inodes_current_limit = default_inodes
+        end
+        set_quota(app, head_blocks_current_limit/1024/1024, head_inodes_current_limit)
+      end
       args = build_base_gear_args(gear)
       args = build_update_cluster_args(options, args)
       result = execute_direct(@@C_CONTROLLER, 'update-cluster', args)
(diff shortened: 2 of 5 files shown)


sallyom:
output:
[fakeapp-sally.dev.rhcloud.com 56be083a97846171ea0000a4]\> quota
Disk quotas for user 56be083a97846171ea0000a4 (uid 5578): 
     Filesystem  blocks   quota   limit   grace   files   quota   limit   grace
     /dev/xvda1 1048576*      0 1048576             203       0   80000  

Successfully moved gear with uuid '56be083a97846171ea0000a4' of app 'fakeapp' from 'ip-172-18-5-113' to 'ip-172-18-2-173'

[root@ip-172-18-5-113 ~]# rhc app show fakeapp --gears quota
Gear                     Cartridges Used Limit
------------------------ ---------- ---- -----
56be083a97846171ea0000a4 php-5.4    1 GB  1 GB

Successfully moved gear with uuid '56be083a97846171ea0000a4' of app 'fakeapp' from 'ip-172-18-2-173' to 'ip-172-18-5-113'

All Successfully complete:
rhc app restart fakeapp, rhc app stop fakeapp, rhc app start fakeapp  

@Miciah, what am I missing? :) Please review

tiwillia:
This seems like it should work. 

Lets [test]

The only worry I have is when `remove_quota` is called during the gear deletion process. At that point, there /should/ be no data on disk used by the gear, so it shouldn't matter there.

sallyom:
@tiwillia please check this out, thanks

sallyom:
[test] pleeeeeeeeese

ironcladlou:
Should the inode quota also be temporarily increased?

tiwillia:
+1 inodes probably also need to be checked and increased.

Currently the changes are permanent. These hanges should probably be temporary, right?

sallyom:
Increasing the inode limit will be np, but first I'll check w/ @a13m  bc the bz doesn't specify a problem with inode quota, only the blocks.  Current workaround script increases inode quota, though so I think I'll just go ahead and do it :)  
Yea, it should be/will be soon temporary :)

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -2029,9 +2029,26 @@ def move_gear(gear, destination_container, destination_district_uuid, change_dis
...
-        quota_files = Integer(gear_quota[6])
+        # if gear is over disk quota limit, increase destination quota by 1% so gear move, start, stop, restart complete
+        # get the base quota blocks and files destination container
+        destination_quota = destination_container.rpc_get_facts_direct(["quota_blocks", "quota_files"])
+        # set destination container quotas
+        quota_blocks = Integer(destination_quota[:quota_blocks])
+        quota_files = Integer(destination_quota[:quota_files])
+        # get current block and file usage from the source container
+        blocks_used = Integer(gear_quota[1])
+        if blocks_used >= quota_blocks
```
a13m: I think we probably need some tolerance here.  It can probably be something like 0.9999, but there are definitely case where usage is a few kilobytes under quota and things still fail due to a combination of rewriting the state file and writing logs.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -2029,9 +2029,26 @@ def move_gear(gear, destination_container, destination_district_uuid, change_dis
...
+          if blocks_used > quota_blocks * 1.01
+            log_debug "Cannot move #{gear.group_instance.gear_size} size gear to #{destination_node_profile}.  Gear exceeds quota limits and quota limit buffer also used up."
+            raise OpenShift::UserException.new("Error moving gear. Cannot move gear to #{destination_node_profile} node profile while gear exceeds profile quota limits.")
+          end
+          log_debug "WARNING: Blocks_used exceeds target disk quota, now using buffer"
+          quota_blocks = blocks_used * 1.01
```
ironcladlou: I'm pretty sure this algorithm can result in unbounded quota increase over repeated calls to `move_gear()` if `blocks_used` increments each time such that `blocks_used` is less than `quota_blocks * 1.01` (where `quota_blocks` is increased with each invocation). In fact, if the quota is applied with a hard limit, I wonder if `blocks_used > quota_blocks * 1.01` would always be false.

Consider the following three move attempts:

```
quota_blocks = 1000 (* 1.01 = 1010)
blocks_used = 1002
quota_blocks = 1012

quota_blocks = 1012 (* 1.01 = 1022)
blocks_used = 1011
quota_blocks = 1021

quota_blocks = 1021 (* 1.01 = 1031)
blocks_used = 1021
quota_blocks = 1031
```

I can imagine how this *might* be desirable since it gives each move operation a chance to progress, but if there's a process that's continuously writing to disk and some system process continuously trying to move the gear, there's now a hole for the process to write unbounded to storage.

Somebody care to verify my analysis?

a13m:
1) Yes, both inode and block quota should be checked / bumped.  inode is not as common to hit, but it happens.
2) Instead of only bumping gears at 100% of quota, I think you need to have some tolerance -- if the gear is already at 99.99% of quota, you should bump it, because chances are is will hit quota during the move, just from log data.
3) for scaled applications, you need to check/bump quota for all gears, because stop/start runs on all gears, and the head gear has to rewrite haproxy.cfg as well.

sallyom:
@a13m and then reset to default limits after the move (that's what I'm working on now)?

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
sallyom: @ironcladlou the max amt the quota limit can be increased is 
1.01 * Integer(destination_quota[:quota_blocks])  
The destination_quota is always based on the default (1048576 blockmax, 8000 inode max for small gears.  The limit will be raised to max of 1059062 then will error rather than give warning.

ironcladlou: >@ironcladlou the max amt the quota limit can be increased is
1.01 * Integer(destination_quota[:quota_blocks])

>The destination_quota is always based on the default (1048576 blockmax, 8000 inode max for small gears. The limit will be raised to max of 1059062 then will error rather than give warning.

That might be my key misunderstanding. When `quota_blocks` is reassigned and then fed to sync, the effective quota for this attempt is NOT persisted or otherwise re-associated with that gear. Thanks, and nevermind!

sallyom:
[test]
@a13m I'm looking at the issue of bumping/checking for all gears in scaled apps.  I notice when moving a non-primary gear that is over limit, it succeeds so long as the head gear is moved first.  Looking into how to best implement your 3) comment above.  Other than that I've implemented all fb.

sallyom:
***SAMPLE OUTPUT FROM A MOVE OF A PRIMARY GEAR THAT IS OVER QUOTA***
App UUID: 56cd9713cde40cbff900001a
Gear UUID: 56cd9713cde40cbff900001a
DEBUG: Source district uuid: 56cd74bfcde40c3b4f000001
DEBUG: Destination district uuid: 869826691282967182966784
WARNING: Quota exceeds 99% of target quota limits, now using buffer.
quota_blocks max with buffer: 1053819, quota_files max with buffer: 80100
DEBUG: Getting existing app 'lastone' status before moving
DEBUG: Gear component 'php-5.4' was running
DEBUG: Stopping existing app cartridge 'php-5.4' before moving
DEBUG: Stopping existing app cartridge 'haproxy-1.4' before moving
DEBUG: Force stopping existing app before moving
DEBUG: Reserved uid '2352' on district: '869826691282967182966784'
DEBUG: Gear platform is 'linux'
DEBUG: Creating new account for gear '56cd9713cde40cbff900001a' on ip-172-18-12-97
DEBUG: Moving content for app 'lastone', gear '56cd9713cde40cbff900001a' to ip-172-18-12-97
Agent pid 23525
unset SSH_AUTH_SOCK;
unset SSH_AGENT_PID;
echo Agent pid 23525 killed;
DEBUG: Moving system components for app 'lastone', gear '56cd9713cde40cbff900001a' to ip-172-18-12-97
Agent pid 23580
unset SSH_AUTH_SOCK;
unset SSH_AGENT_PID;
echo Agent pid 23580 killed;
DEBUG: Starting cartridge 'haproxy-1.4' in 'lastone' after move on ip-172-18-12-97
DEBUG: Starting cartridge 'php-5.4' in 'lastone' after move on ip-172-18-12-97
DEBUG: Fixing DNS and mongo for gear '56cd9713cde40cbff900001a' after move
DEBUG: Changing server identity of '56cd9713cde40cbff900001a' from 'ip-172-18-10-186' to 'ip-172-18-12-97'
DEBUG: Deconfiguring old app 'lastone' on ip-172-18-10-186 after move
WARNING:  Gear was moved successfully but quota limits for small node profile exceeded.

sallyom:
@a13 @tiwillia for your review.
Please see my comments here: https://github.com/sallyom/origin-server/blob/bzgearexceedquota/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb#L1771  
I will continue testing to be sure my comment holds true.  Thanks

sallyom:
@tiwillia @a13m  
1) increased max to 1.2 times original and each time increments by 1% or 50 inodes
2) for gear moves, will bump if quota > 95% limits, for update-cluster will bump if quota > 99% limits
    less for update-cluster to minimize double-bumps
looking good? :)

tiwillia:
@sallyom this LGTM.

This looks ready to merge. Should we wait for @a13m 's blessing?

broker-util/oo-admin-move
```diff
@@ -156,6 +156,8 @@ gear_uuid.split(',').each do |uuid|
         puts ne.message
       end
       rc |= 1
+    rescue OpenShift::QuotaExceededException => qe
+      puts qe.message
```
a13m: This will break our json output mode.  Look at how we are handling OpenShift::UserException in the subsequent few lines.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -1768,6 +1769,35 @@ def build_update_cluster_args(options, args)
...
+      app = gear.application
+      head_gear_quota = get_quota(app)
+      # set head_gear quotas if necessary
+      # will be necessary if trying to move a non-primary gear before bumping
+      # an over-the-quota-limit head gear
+      head_blocks_current_limit = Integer(head_gear_quota[3])
+      head_blocks_used = Integer(head_gear_quota[1])
+      head_blocks_increment = head_blocks_current_limit * 1.01
+      head_quota_blocks_buffer = head_blocks_current_limit * 1.20
+
```
a13m: these hard-coded numbers are fine with me, but I don't know whether they will be okay for an OSE customer.  you should probably run that by someone more familiar with OSE deployments; they might want something configurable.

sallyom:
@tiwillia @a13m   
I made the buffer configurable. 
Also, if usage that was previously over-limit is decreased under default quota limits the limits will be reset to default in next move/update-cluster.
Increments of 1%, 10 inodes with warnings during move and during update-cluster to max of 
(default * configured multiplier)...hopefully it will pass all tests and be ready for merge now. : )

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -34,7 +34,9 @@ class MCollectiveApplicationContainerProxy < OpenShift::ApplicationContainerProx
     # * id: string - a unique app identifier
     # * district: <type> - a classifier for app placement
     #
-    def initialize(id, district=nil)
+    def initialize(id, district=nil, blocks_multiplier=Rails.configuration.msg_broker[:quota_blocks_buffer], inodes_multiplier=Rails.configuration.msg_broker[:quota_inodes_buffer])
```
tiwillia: Not an actual issue, but style-wise this is probably incorrect. The line is far too long, perhaps set blocks_multiplier et al to `nil` by default and set it to the configuration value on the next line if it is `nil`.

tiwillia: I guess there is precedent for doing this in the `node-util/conf/watchman/plugins.d/env_plugin.rb`'s initialize method, so its up to you whether to leave it be or not.

tiwillia: Well, what will `Rails.configuration.msg_broker[:quota_blocks_buffer]` return if the configuration doesn't exist?

tiwillia:
There is going to be a bit of confusion about having this configuration on both the broker and the node. With so many nodes in the environment, it could be possible that an administrator sets the node's configuration different from that set on the broker.

Would the only result of this be that the warning that comes from the node would be shown at incorrect quota usage?

sallyom:
@tiwillia I made the style change you suggested, thanks.   I'm looking into removing that node.conf change now, I don' think it's necessary.

node/test/node_functional/node_func_test.rb
```diff
@@ -15,6 +15,7 @@
 #++
 require_relative '../test_helper'
 require 'openshift-origin-node/model/node'
+require 'openshift-origin-common'
```
tiwillia: Is this actually needed?

sallyom: nope, thanks for catching that :)  I removed that line.

tiwillia:
Alrighty, I think this is ready to [merge]

Thanks for putting this together @sallyom !

tiwillia:
Merge failure appears to be unrelated. Lets try this again and see if we get different results [merge]

PR #6368, opened 2016-02-26
Updates _mandir to properly add node man pages

thrasher-redhat:
Bug 1031796
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1031796

Updates _mandir to _root_mandir in the node .spec file.  The node uses SCL, so
_mandir would incorrectly put the man pages into
/opt/rh/ruby193/root/user/share/man instead of into /usr/share/man, which would
require users to enable SCL to update the $MANPATH.  The node
will now add the man pages to the correct location.

Specifically the man pages for oo-admin-ctl-tc and oo-admin-ctl-iptables-port-proxy.


Miciah:
Nice! Please [merge]!

PR #6369, opened 2016-02-27
Remove double warning msg when adding cartridges

sallyom:
bz1169690
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1169690

When adding cartridges to an app that is approaching quota limits,
the "Warning: Gear <uuid> is using <percent> of disk quota" is printed twice.
This PR removes the report_quota when installing a cartridge, so the
quota is only reported once after the cartridge is installed.

### DIFF
--- a/plugins/msg-node/mcollective/src/openshift.rb
+++ b/plugins/msg-node/mcollective/src/openshift.rb
@@ -157,7 +157,9 @@ module MCollective
         exitcode, output, addtl_params = execute_action(action, args)
 
         if args['--with-container-uuid']
-          report_quota(output, args['--with-container-uuid'])
+          if action != "configure"
+            report_quota(output, args['--with-container-uuid'])
+          end
           if report_resource(output, args['--with-app-name'])
             exitcode = 222
           end


sallyom:
[test]

plugins/msg-node/mcollective/src/openshift.rb
```diff
@@ -157,7 +157,12 @@ def cartridge_do_action
         exitcode, output, addtl_params = execute_action(action, args)
 
         if args['--with-container-uuid']
-          report_quota(output, args['--with-container-uuid'])
+          if action.to_s.include?("post-configure") || action.to_s.include?("deconfigure")
+            report_quota(output, args['--with-container-uuid'])
+          else
+            # do not print quota warning (output) during action "configure", to avoid double warning msg
+            report_quota('', args['--with-container-uuid'])
```
tiwillia: If we aren't going to do anything with the output, why run `report_quota` at all? It doesn't appear to do anything other than generate output, so it seems needless to run it in this case.

Also, to ensure that quota is reported with everything but the 'configure' step, why not use `if action.to_s != "configure"` ?

sallyom:
@tiwillia all set? thanks

node/test/unit/openshift_test.rb
```diff
@@ -159,7 +159,6 @@ def test_cartridge_do_action_quota
     @agent.expects(:request).returns(request).times(3)
 
     @agent.expects(:execute_action).with('test_agent', any_parameters).returns [0, 'test output', nil]
-    @agent.expects(:report_quota).with(instance_of(String), 'uuid')
```
tiwillia: Could you explain why this was removed? It looks like the test uses the action 'test_agent', which should match your conditional and report quota.

sallyom:
@tiwillia fixed up, thank you

tiwillia:
LGTM. In testing, I can see that only one warning message is seen when adding a cartridge.

[merge]

PR #6370, opened 2016-02-29
oo-accept-node: check-quotas: fix mnt opts parsing

Miciah:
Improve the parsing of mount options from the `mount` command's output by properly splitting on newlines and then removing the parentheses around the mount options.

Before this commit, the pattern match `/loop=([^,]+)/` on the mount options was capturing the right parenthesis and newline from the `mount` command's output if the loop option was the last mount option.

This commit is related to bug 1265811.

https://bugzilla.redhat.com/show_bug.cgi?id=1265811


Miciah:
Please [test]!

tiwillia:
LGTM!

[merge]

PR #6373, opened 2016-03-23
Bug 1313395 - Forward several headers in websocket upgrade requests

a13m:


tiwillia:
This LGTM but I don't know much about the subject matter. @Miciah are you familiar enough with this to review?

node-proxy/lib/proxy/ProxyServer.js
```diff
@@ -453,10 +453,21 @@ function finish_websocket(upg_reqhost, proxy_server, ws) {
...
+  zheaders.headers['user-agent'] = upgrade_req.headers['user-agent'];
+
+  /*  Set X-Forwarded-For HTTP extension header.  */
+  var xff = upgrade_req.connection.remoteAddress  ||
+            upgrade_req.socket.remoteAddress;
+  zheaders.headers['X-Forwarded-For'] = xff;
+
+  /*  Set X-Client-IP HTTP extension header.      */
+  zheaders.headers['X-Client-IP'] = xff;
+
```
Miciah: Does case matter here? We're sometimes using lower-case, sometimes using capitals.

HTTP is supposed to be case-insensitive, the documentation for the http module indicates Node.js normalises headers to lower-case when they are received, and as far as I can tell the http module's accessors do the same, so lower-case seems to be preferred in Node.js (although some examples in the documentation use capitals).  However, the documentation for the ws module doesn't say anything at all about case, and we are constructing and passing in the headers object ourselves rather than using any sort of accessor, so I'm wary of making assumptions.

@bparees, does this fall within your expertise?

bparees: not particularly, but i'd second your belief that case is irrelevant, so it's basically aesthetics.

Miciah:

PR #6374, opened 2016-03-23
Allow unidling to ignore configurable user agent strings

tiwillia:
When undiling, it may be necessary to ignore certain user agent strings to avoid unidling gears that do not need to be unidled. For example, a request from a search engine spider should not cause the application to unidle. This change adds the ability to configure a list of strings that will be checked against the user agent string of each request and will cause the unilder to ignore the request if there is a match.

Bug 1305166
https://bugzilla.redhat.com/show_bug.cgi?id=1305166

### DIFF
--- a/node-util/sbin/oo-last-access
+++ b/node-util/sbin/oo-last-access
@@ -27,7 +27,13 @@ class LastAccessUpdater
     @last_access_data = Hash.new
     @addrs_ignore = Array.new
     @hosts_ignore = ["localhost"]
+    @useragent_ignore = []
   end
+
+  def load_useragent_ignore
+    ignore_list = OpenShift::Config.new.get("UNIDLER_IGNORE_LIST") || ""
+    @useragent_ignore = ignore_list.split(',')
+   end
 
   def load_addrs_ignore
     matchaddr = Regexp.new('inet6? (.*)\/[0-9]')
@@ -115,7 +121,8 @@ class LastAccessUpdater
     # Doing the file read with awk is 10x faster than doing it inside
     # Ruby and we desperately need speed here.
     ignore_cond = [@hosts_ignore.map { |a| "$2 != \"#{a}\"" },
-                   @addrs_ignore.map { |a| "$1 != \"#{a}\"" }].flatten.join(' && ')
+                   @addrs_ignore.map { |a| "$1 != \"#{a}\"" },
+                  "$_ !~ /#{ @useragent_ignore.join('|') }/" ].flatten.join(' && ')
     awk_cmd = %Q[{ if ( #{ignore_cond} ) h[$2]=$5" "$6 } END{ for ( i in h ) { printf("%s=%s\\n",i,h[i]) } }]
     # Pack separate log files one at a time so we can enforce the rule
     # that most recent timestamp wins.
@@ -207,6 +214,7 @@ single_instance do
   logfiles = [ENV["APACHE_ACCESS_LOG"], ENV["NODE_WEB_PROXY_ACCESS_LOG"], ENV["NODE_WEB_PROXY_WEBSOCKETS_LOG"]]
   l = LastAccessUpdater.new(ENV["GEAR_BASE_DIR"], logfiles)
   l.load_addrs_ignore
+  l.load_useragent_ignore
   l.load_gears_data
   l.add_alias_data
   gears_found_count = l.parse_log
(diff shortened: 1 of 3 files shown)


tiwillia:
[test]

tiwillia:
@Miciah would you mind giving this a look when you have time?

Do you think we should do anything with the UNIDLER_IGNORE_LIST value before passing it to awk?

tiwillia:
Very weird failures... lets re[test]

tiwillia:
I reproduced this test failure in an ec2 instance. Simply restarting the broker resolved the issue entirely and allowed rhc-accept-env to pass. Not sure exactly what is causing it.

Miciah:
Please [test] again.

node-util/sbin/oo-last-access
```diff
@@ -27,8 +27,14 @@ class LastAccessUpdater
     @last_access_data = Hash.new
     @addrs_ignore = Array.new
     @hosts_ignore = ["localhost"]
+    @useragent_ignore = []
   end
 
+  def load_useragent_ignore
+    ignore_list = OpenShift::Config.new.get("UNIDLER_IGNORE_LIST") || ""
```
Miciah: Instead of using `|| ""`, you can specify a default value to `get`: `OpenShift::Config.new.get("UNIDLER_IGNORE_LIST", "")`

node-util/www/html/restorer.php
```diff
@@ -2,6 +2,24 @@
...
+        $useragent_ignore_value = array();
+        preg_match('/UNIDLER_IGNORE_LIST=[\'"]*([^"\']+)[\'"]*/', $conf_line, $useragent_ignore_value);
+        $useragent_ignore_list = explode(",", $useragent_ignore_value[1]);
+        break;
+      }
+    }
+    // If the user agent matches one in the list, move on.
+    foreach ($useragent_ignore_list as $ignore) {
+      if (preg_match("/" . $ignore . "/", $_SERVER["HTTP_USER_AGENT"])) {
+        header('HTTP/1.0 403 Forbidden');
```
Miciah: Should we log something?

node-util/sbin/oo-last-access
```diff
@@ -115,7 +121,8 @@ class LastAccessUpdater
     # Doing the file read with awk is 10x faster than doing it inside
     # Ruby and we desperately need speed here.
     ignore_cond = [@hosts_ignore.map { |a| "$2 != \"#{a}\"" },
-                   @addrs_ignore.map { |a| "$1 != \"#{a}\"" }].flatten.join(' && ')
+                   @addrs_ignore.map { |a| "$1 != \"#{a}\"" },
+                  "$_ !~ /#{ @useragent_ignore.join('|') }/" ].flatten.join(' && ')
```
Miciah: This will break if `UNIDLER_IGNORE_LIST` contains slashes, and User-Agent strings can (and usually do) include slashes.  This will also interpret special regex characters (".", "*", "+", "?", "(", ")", "[", "]", "|", "\").  We should escape `@useragent_ignore`.

Could someone create an application or domain that contained something in `UNIDLER_IGNORE_LIST`? We probably shouldn't use `$_`.

node-util/www/html/restorer.php
tiwillia: I thought about that too. Is there any reason to? I guess we could log that a request was ignored for app with <uuid> with user-agent string <user-agent string>. Would that be useful?

Miciah: I would think yes because it should ease diagnosis should a user come complaining about mysterious 403s when we misconfigure `UNIDLER_IGNORE_LIST` or some client program or library happens to use some blacklisted User-Agent string.  Generally, I prefer more logging over less—but I don't deal with logging at the scale that ops does.  I suppose this is a question for @a13m.

node-util/sbin/oo-last-access
Miciah: BTW, what happens when `@useragent_ignore` is empty? Looks like it'll boil down to `$_ !~ //`, which will always be false, so we'lll always ignore the log entry.

tiwillia: I see some value in allowing regex and defining the option as such in the node.conf.

Why exactly shouldn't we `$_`?

Miciah: Because it matches the entire line, not just the User-Agent string.

Also, I just realised, it's undocumented and non-standard.

tiwillia:
After having a discussion with @a13m, it was realized that this change is not needed. Closing this PR for now.

PR #6378, opened 2016-04-13
Update manifest.yml

vandat6x:


Miciah:
Is there a problem that this PR is intended to fix, or was it created by accident?

PR #6379, opened 2016-04-15
Call setter to initialize watchman configuration valriables

nak3:


nak3:
Due to this issue, `Rational(seconds, 86400)` is not called and some plugins doesn't start from the second time.

e.g retry_delay is critical.
https://github.com/openshift/origin-server/blob/73197c277b9ce9044be67a4e23b9951d120bc663/node-util/sbin/oo-watchman#L110

PR #6380, opened 2016-04-15
Bug 1327714: MySQL cartridge removes logs on start

dinhxuanvu:
The "start" function of start script of the MySQL cartridge removes the
mysql.log which is supposed to save some disk space. However, since the
log file is removed, it's very difficult, if not impossible, for users
to debug any issues with MySQL database.

Since the log file removal is unnecessary, the code that is responsible
for that is now removed. As a result, the mysql.log file will remain
after the "start" function in control script is called.

Bug: 1327714
Link: https://bugzilla.redhat.com/show_bug.cgi?id=1327714

### DIFF
--- a/cartridges/openshift-origin-cartridge-mysql/bin/control
+++ b/cartridges/openshift-origin-cartridge-mysql/bin/control
@@ -41,7 +41,6 @@ function is_running {
 function start {
   if ! is_running; then
     echo "Starting MySQL ${OPENSHIFT_MYSQL_VERSION} cartridge"
-    rm -f $OPENSHIFT_LOG_DIR/mysql*.log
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
[test] @a13m @Miciah Please review. Thanks.

dinhxuanvu:
re-[test]

dinhxuanvu:
Everything looks good. @Miciah or @tiwillia waiting for your blessing :)...

tiwillia:
Since the `tidy` control function handles deleting the mysql log file this LGTM. Users will still be able to delete the log without manually deleting it in the gear when it is taking up too much space.

[merge]

PR #6381, opened 2016-04-19
cron_runjobs.sh: Feature Enhancement for syslog tagging

sferich888:
Bug 1217403
Bugzilla link https://bugzilla.redhat.com/show_bug.cgi?id=1217403

Adding the ability to filter/tag cron messages with syslog

### DIFF
--- a/cartridges/openshift-origin-cartridge-cron/bin/cron_runjobs.sh
+++ b/cartridges/openshift-origin-cartridge-cron/bin/cron_runjobs.sh
@@ -37,7 +37,7 @@ CART_CONF_DIR=$OPENSHIFT_CRON_DIR/configuration
 function log_message() {
    msg=${1-""}
    [ -z "$msg" ]  &&  return 0
-   logger -i -s "user-cron-jobs" -p user.info "`date`: $msg"
+   logger -i -s "user-cron-jobs" -p user.info -t "cron_sys_log:" "`date`: $msg"
 }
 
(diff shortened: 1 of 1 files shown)


sferich888:
The rsyslog configuration file also needs to be modified for this to work. 
```
:syslogtag, isequal, "cron_sys_log:" /var/log/ABC.log 
```
This needs to be added to the config, to specify where you wan to send the cron logs. Documentation on this should be updated with https://bugzilla.redhat.com/show_bug.cgi?id=1328495

sferich888:
[test]

Miciah:
@sferich888, thanks! Looks good.  Do we want to configure rsyslog to log system-level cron logging output to a separate file as you describe (maybe `/var/log/openshift/node/cron.log`) by default? If so, I will make the PR for the installation script.

sferich888:
@Michiah your call I honestly think letting an admin pick the location is a feature but I see the benefit of defining a default at install time.

Miciah:
Yeah, the administrator can always modify `/etc/rsyslog.d/imuxsock-and-openshift-gears.conf` after installation to comment out the line or change the path (which will be easier to do if the `:syslogtag` stanza is already there).

Miciah:
The tests got a lot of timeouts—most likely we got a dud EC2 instance.  Please [test] again!

Miciah:

PR #6382, opened 2016-04-24
ensure that gears are moved to nodes with adequate free disk space

sallyom:
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1122084
bz 1122084
If a gear is moved to a node and that node has no free space left
after the move, the node becomes unusable.
This PR ensures adequate free space + a buffer of free space on
destination nodes when moving gears.

### DIFF
--- a/node/lib/openshift-origin-node/model/node.rb
+++ b/node/lib/openshift-origin-node/model/node.rb
@@ -347,6 +347,8 @@ module OpenShift
         res['gears_started_count'] = 0
         res['gears_deploying_count'] = 0
         res['gears_unknown_count'] = 0
+        res['node_disk_free'] = 0
+        res['node_total_size'] = 0
         OpenShift::Runtime::ApplicationContainer.all(nil, false).each do |app|
           # res['git_repos_count'] += 1 if ApplicationRepository.new(app).exists?
           res['gears_total_count'] += 1
@@ -365,6 +367,15 @@ module OpenShift
             res['gears_unknown_count'] += 1
           end
         end
+
+        # fact for available disk space on a node, considered in gear moves to ensure minimum buffer of free space on nodes
+        mountpoint = self.get_gear_mountpoint
+        df_cmd = "df -aP #{mountpoint} | awk -F' ' 'NR == 2 {print $2,$4}'"
+        df_output, stderr, rc = Utils.oo_spawn(df_cmd)
+        raise NodeCommandException.new "Error: #{stderr} executing command #{df_cmd}" unless rc == 0
+        df_total_size, df_free = df_output.split
+        res['node_total_size'] = df_total_size.to_i
+        res['node_disk_free'] = df_free.to_i
 
         # consider a gear active unless explicitly not
         res['gears_active_count'] = res['gears_total_count'] - res['gears_idled_count'] - res['gears_stopped_count']
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -788,6 +788,34 @@ module OpenShift
       rpc_get_fact_direct('quota_files').to_i
     end
 
+    # <<accessor>>
+    # Get the available disk space of a Node
+    #
+    # RETURNS:
+    # * Integer: the available disk space of a node in blocks
+    #
+    # NOTES:
+    # * method on Node
+    # * calls rpc_get_fact_direct
+    #
+    def get_node_disk_free
+      rpc_get_fact_direct('node_disk_free').to_i
+    end
+
+    # <<accessor>>
+    # Get the total disk space of a Node
+    #
+    # RETURNS:
+    # * Integer: the total disk space of a node in blocks
+    #
+    # NOTES:
+    # * method on Node
+    # * calls rpc_get_fact_direct
+    #
+    def get_node_total_size
+      rpc_get_fact_direct('node_total_size').to_i
+    end
+
     #
     # Add a component to an existing gear on the node
     #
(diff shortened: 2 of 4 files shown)


sallyom:
WIP
[test] please

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -2375,7 +2389,16 @@ def rsync_destination_container(gear, destination_container, destination_distric
       reply = ResultIO.new
       source_container = gear.get_proxy
       platform = gear.group_instance.platform
+      gear_quota = get_quota(gear)
       log_debug "DEBUG: Gear platform is '#{platform}'"
+      destination_available_space = destination_container.get_node_disk_free
+
+      # check here to make sure addition of gear to destination_container
+      # will not result in > 98% of available free space on destination_container
+      if gear_quota[1].to_i > 0.98 * destination_available_space
```
Miciah: It's not obvious from the name what `gear_quota[1]` is, and you use it twice, so a descriptively named local variable (e.g., `gear_blocks_used`) might be in order.

node/lib/openshift-origin-node/model/node.rb
```diff
@@ -366,6 +366,13 @@ def self.node_utilization
           end
         end
 
+        # fact for available disk space on a node, considered in gear moves to ensure minimum buffer of free space on nodes
+        mountpoint = self.get_gear_mountpoint
+        cmd = "df #{mountpoint} | awk -F' ' 'NR == 2 {print $4}'"
+        df, stderr, rc = Utils.oo_spawn(cmd)
+        raise NodeCommandException.new "Error: #{stderr} executing command #{cmd}" unless rc == 0
+        res['node_disk_free'] = df
```
Miciah: `df` has a newline character at this point (you can see this if you view `/opt/rh/ruby193/root/etc/mcollective/facts.yaml`); use `df.chomp` instead to remove it.  Alternatively, use `to_i` here (`to_i` ignores whitespace) and store the fact as an integer instead of a string.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -788,6 +788,20 @@ def get_quota_files
       rpc_get_fact_direct('quota_files').to_i
     end
 
+    # <<accessor>>
+    # Get the available disk space of a Node
+    #
+    # RETURNS:
+    # * Integer: the available disk space of a node
```
Miciah: Please specify the unit: "available disk space of a node in blocks".

sallyom:
@Miciah, I've addressed your feedback. Thanks!!

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -2375,7 +2389,16 @@ def rsync_destination_container(gear, destination_container, destination_distric
       reply = ResultIO.new
       source_container = gear.get_proxy
       platform = gear.group_instance.platform
+      source_used_blocks = get_quota(gear)
```
Miciah: I think it would make sense to assign `source_used_blocks` the value of `get_quota(gear)[1]` instead of `get_quota(gear)`.  It would be a little more prudent and explicit to do something like this:

```ruby
      quota = get_quota(gear)
      source_used_blocks = quota[1] unless quota.nil?
```

But I don't feel strongly about it.

sallyom: I feel strongly that you are correct :) so I made the change.

sallyom:
@a13m can you please review?  Buffer too much/too little?  Thanks.

Miciah:
openshift-bot, please run the [extended:node] [test]s.

node/lib/openshift-origin-node/model/node.rb
```diff
@@ -366,6 +367,13 @@ def self.node_utilization
           end
         end
 
+        # fact for available disk space on a node, considered in gear moves to ensure minimum buffer of free space on nodes
+        mountpoint = self.get_gear_mountpoint
+        cmd = "df -a #{mountpoint} | awk -F' ' 'NR == 2 {print $4}'"
```
a13m: This doesn't work when the output spans lines.  You need -P

a13m:
Thanks for working on this.  I think I would have preferred a check like 

```
(DESTINATION_AVAILABLE_SPACE - SOURCE_GEAR_QUOTA) / DESTINATION_TOTAL_SPACE > 0.05
```

In other words, make sure that moving the gear to this node does not cause the disk usage to exceed 95% of its capacity (possibly even 90%).  The way you have it written, it seems like we could have 1.03 GB free and then still write a 1 GB gear to the node.  We don't like to cut things that close!
I would be okay with making that threshold configurable in case OSE customers want something different.

node/lib/openshift-origin-node/model/node.rb
```diff
@@ -366,6 +368,17 @@ def self.node_utilization
...
         end
 
+        # fact for available disk space on a node, considered in gear moves to ensure minimum buffer of free space on nodes
+        mountpoint = self.get_gear_mountpoint
+        cmd_free = "df -aP #{mountpoint} | awk -F' ' 'NR == 2 {print $4}'"
+        cmd_total_size = "df -aP #{mountpoint} | awk -F' ' 'NR == 2 {print $2}'"
+        df_free, stderr, rc = Utils.oo_spawn(cmd_free)
+        raise NodeCommandException.new "Error: #{stderr} executing command #{cmd_free}" unless rc == 0
+        df_total_size, stderr, rc = Utils.oo_spawn(cmd_total_size)
+        raise NodeCommandException.new "Error: #{stderr} executing command #{cmd_total_size}" unless rc == 0
```
Miciah: What do you think about doing something like the following?
```ruby
        df_cmd = "df -aP #{mountpoint} | awk -F' ' 'NR == 2 {print $2 \" \" $4}'"
        df_output, stderr, rc = Utils.oo_spawn(df_cmd)
        raise NodeCommandException.new "Error: #{stderr} executing command #{df_cmd}" unless rc == 0
        df_total_size, df_free = df_output.split
```

Miciah:
I see no need to make the threshold configurable unless a customer specifically requests it.  95% seems like a thoroughly reasonable threshold.

Miciah:
@sallyom, the "Online Test Results: Waiting for stable build of 'libra_ami'" message means the image that we use for running tests is currently failing tests and needs to be rebuilt (possibly after fixing whatever problem is causing it to fail tests).

sallyom:
@Miciah, awesome, thank you.

sallyom:
[test] pleeeeeeese, Ms. openshift-bot

sallyom:
@Miciah review please thanks

Miciah:

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -2375,7 +2403,18 @@ def rsync_destination_container(gear, destination_container, destination_distric
...
       platform = gear.group_instance.platform
+      quota = get_quota(gear)
+      source_used_blocks = quota[1] unless quota.nil?
       log_debug "DEBUG: Gear platform is '#{platform}'"
+      destination_avail_space = destination_container.get_node_disk_free
+      destination_total_space = destination_container.get_node_total_size
+
+      # check here to make sure addition of gear to destination_container
+      # will not result in > 95% full destination_container
+      if (destination_avail_space - source_used_blocks.to_i)/destination_total_space > 0.05
```
jianlinliu: Here ">" is right? I feel like it should be "<".

PR #6383, opened 2016-04-27
Tomcat does not properly parse spaces in JVM parameters

dinhxuanvu:
Tomcat script doesn't correct handle the white places in the parameters passing
via env variable JAVA_OPTS_EXT which causes the JVM fails to parse the parameters
correctly during execution.

The tomcat bash script is modified to handle the white spaces and ensured the space
is not ignored by adding the extra double quotes around the "JAVA_OPTS_EXT".

Bug: 1282852
Link: https://bugzilla.redhat.com/show_bug.cgi?id=1282852

### DIFF
--- a/cartridges/openshift-origin-cartridge-jbossews/bin/tomcat
+++ b/cartridges/openshift-origin-cartridge-jbossews/bin/tomcat
@@ -32,16 +32,12 @@ memory_options="-Xms40m -Xmx${max_heap}m -XX:MaxPermSize=${max_permgen}m -XX:+Ag
 # Specify options to pass to the Java VM.
 #
 if [ -z "$JAVA_OPTS" ]; then
-   JAVA_OPTS="$memory_options -DOPENSHIFT_APP_UUID=${OPENSHIFT_APP_UUID} -Djava.util.logging.config.file=$CATALINA_BASE/conf/logging.properties -Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager -Djava.net.preferIPv4Stack=true -Dfile.encoding=UTF-8 -Djava.net.preferIPv4Stack=true"                                                                                                                                    
+   JAVA_OPTS="$memory_options -DOPENSHIFT_APP_UUID=${OPENSHIFT_APP_UUID} -Djava.util.logging.config.file=$CATALINA_BASE/conf/logging.properties -Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager -Djava.net.preferIPv4Stack=true -Dfile.encoding=UTF-8 -Djava.net.preferIPv4Stack=true"
 
    if marker_present "enable_jpda"; then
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
[test]

dinhxuanvu:
@Miciah Please review. Thanks :)

cartridges/openshift-origin-cartridge-jbossews/bin/tomcat
```diff
@@ -87,42 +83,84 @@ EOF
...
-    -Dcatalina.home="$CATALINA_HOME" \
-    -Djava.endorsed.dirs="$JAVA_ENDORSED_DIRS" \
-    -Djava.io.tmpdir="$CATALINA_TMPDIR" \
-    org.apache.catalina.startup.Bootstrap stop 2>&1
-elif [ "$1" = "version" ]; then
-  ${JAVACMD} -classpath ${CATALINA_HOME}/lib/catalina.jar \
-    org.apache.catalina.util.ServerInfo
+if [ -n "$JAVA_OPTS_EXT" ]; then
+  if [ "$1" = "start" ]; then
+      ${JAVACMD} $JAVA_OPTS "$JAVA_OPTS_EXT" $LOGGING_CONFIG $CATALINA_OPTS \
```
Miciah: You could avoid all the duplication by using `${JAVA_OPTS_EXT:+"$JAVA_OPTS_EXT"}`. `${foo:+bar}` substitutes `bar` iff `$foo` is set and is non-empty, so we will get quoted `$JAVA_OPTS_EXT` if it is present and nothing otherwise.

Miciah: Will quoting `$JAVA_OPTS_EXT` fix the general problem? The Bugzilla report uses the following command in the steps to reproduce:
```console
$ rhc env set JAVA_OPTS_EXT="-Dworking.dir=app-root/data/test dir" -a APP_NAME
```
After this change, using this first example, you will indeed pass `-Dworking.dir=app-root/data/test dir` as a single argument to `${JAVACMD}`.  However, consider a second example:
```console
$ rhc env set JAVA_OPTS_EXT="-Dfoo=bar -Dbaz=quux" -a APP_NAME
```
Now we will pass `-Dfoo=bar -Dbaz=quux` as a single argument to `${JAVACMD}`.  Is this second example something we should support? Don't we need to split `$JAVA_OPTS_EXT` in the second example?

dinhxuanvu:
@Miciah Fixed and tested :D... Please review again and bless it if you wish :D...

Miciah:
openshift-bot, please [test]! [extended:cartridge]

dinhxuanvu:
re-[test]

Miciah:
Please [test] again!

dinhxuanvu:
Everything looks good. Until you have something to add @Miciah, please bless it with a merge :)

Miciah:

PR #6384, opened 2016-04-27
Openshift doesn't accept app aliases containing an underscore

dinhxuanvu:
Currently, application aliases with underscore(s) are not allowed due
to regex restriction during alias verification
(controller/app/models/application.rb validate_alias(fqdn)).

Having underscore(s) in URL is generally not recommended but it's not
technically wrong to have. So, the regex for validate_alias method
is modified to accept underscore(s).

Bug: 1329915
Link: https://bugzilla.redhat.com/show_bug.cgi?id=1329915

### DIFF
--- a/controller/app/models/application.rb
+++ b/controller/app/models/application.rb
@@ -1376,7 +1376,7 @@ class Application
     fqdn.downcase!
     return false if fqdn =~ /^\d+\.\d+\.\d+\.\d+$/
     return false if fqdn =~ /\A[\S]+(\.(json|xml|yml|yaml|html|xhtml))\z/
-    return false if not fqdn =~ /\A[a-z0-9]+([\.]?[\-a-z0-9]+)+\z/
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
[test]
@tiwillia Please review. Thanks :)

tiwillia:
LGTM, but...

[ RFC 1123, section 2.1](https://tools.ietf.org/html/rfc1123#page-13) refers to [RFC 952](https://tools.ietf.org/html/rfc952) to determine proper hostname/subdomain syntax. The RFC states that the name should match the following: `<let>[*[<let-or-digit-or-hyphen>]<let-or-digit>]`. Underscores do not look to be allowed.

cc @Miciah

dinhxuanvu:
As @tiwillia comment above, the underscore should not be allowed in alias. So this PR and the bug associated are closed.
Thanks all!

PR #6388, opened 2016-05-20
Fixing short-form optparse options

thrasher-redhat:
Bug 1311730
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1311730

Errors were being seen when command line tools using the 'optparse' libary.
Using long-form only options with a single dash was resulting in odd returns,
but no errors.  This is because optparse automatically infers a short-form
option for long-form options that have a unique starting letter.  So '--ids' has
an implied short-form of '-i'.  This meant that running a command with '-ids'
would actually use '-i' and pass in "ds" as an ID value.  This commit takes
several cases where this happens and explicitely defines the short-form options
in the relevant scripts and man pages, which should reduce confusion on why
certain options may not error out as expected.

Also updates an old man page to reflect the current options.


tiwillia:
LGTM

Since we don't have any automated tests that cover these scripts (and most of your changes just result in help page changes), I tested these myself. Did not find any problem with the modified admin scripts and saw the options in the help pages where expected.

[merge]

PR #6392, opened 2016-06-08
Added documentation on `--change_region`

brs17:
While the implementation of `--change_region` exisited in oo-admin-move, there had been no mention of the option in the output of either `oo-admin-move --help` nor in the man page.


a13m:
@brs17  This was intentionally left undocumented when it was implemented.  If it is added to docs, it's needs to have strong admonitions about being unsupported, not recommended for scaled apps, etc.

brs17:
@a13m Made the PR because I was in a situation where I needed to use it but had very little guidance. Did not realize how little it was used/supported.

PR #6393, opened 2016-06-10
Publish  Routing Info During 'oo-admin-move'

sallyom:
bz 1274852

Update LB when head gear is moved.  Ensure that routing information is updated during gear moves.

### DIFF
--- a/controller/app/models/gear.rb
+++ b/controller/app/models/gear.rb
@@ -132,6 +132,12 @@ class Gear
     result_io
   end
 
+  def unpublish_routing_info
+    self.port_interfaces.each do |pi|
+      pi.unpublish_endpoint(self.application)
+    end
+  end
+
   def publish_routing_info
     self.port_interfaces.each do |pi|
       pi.publish_endpoint(self.application)
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -1932,6 +1932,8 @@ module OpenShift
       ensure
         dns.close
       end
+      log_debug "DEBUG: Updating routing information for gear '#{gear.uuid}' after move"
+      gear.publish_routing_info
       reply
     end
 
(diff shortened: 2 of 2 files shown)


sallyom:
@Miciah please review, thank you.

sallyom:
[test] please

Miciah:
Nice! But what about telling the load balancer to remove the old endpoints?

sallyom:
@Miciah, I added an 'unpublish' call in move_gear_pre.  I believe this is the right place for it.  Works in my test env.  Ready for another review.

plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
```diff
@@ -1986,6 +1988,9 @@ def move_gear_pre(gear, state_map)
           reply.append source_container.force_stop(gear)
         end
       end
+
+      log_debug "DEBUG: Unpublishing routing information for gear '#{gear.uuid}'"
+      gear.unpublish_routing_info
```
Miciah: We should probably do this up front, right before line 1962, so that we tell the load balancer to stop routing to the gear *before* we stop the gear.

controller/app/models/gear.rb
```diff
@@ -132,6 +132,12 @@ def destroy_gear(keep_uid=false, is_group_rollback=false)
     result_io
   end
 
+  def unpublish_routing_info
+    self.port_interfaces.each do |pi|
+      pi.upublish_endpoint(self.application)
```
Miciah: Type: `upublish`.

Miciah:
Looks good but for 2 comments.  To verify this change, we should make sure we see the expected notifications in `routing-daemon.log` and that nginx (assuming you are testing with it) is configured properly (with the old endpoint replaced by the new endpoint) and routes properly after moving a gear.

sallyom:
@Miciah, expected notifications confirmed in routing-daemon.log and endpoints in '/opt/rh/nginx16/root/etc/nginx/conf.d/pool_ose_lbtest_sally_80.conf' (in my env) are also updated.

Miciah:

Miciah:
Thanks, openshift-bot!

PR #6397, opened 2016-06-24
Bug 1311722 Modify docs for oo-admin-cartridge -a erase command

dinhxuanvu:
This PR modifies the manpage and the output of --help option to
highlight the fact that 'oo-admin-cartridge -a erase' will remove all
version of given cartridge version despite users may only specify
only one version in '-v' argument.

Bug: 1311722
Link: https://bugzilla.redhat.com/show_bug.cgi?id=1311722


dinhxuanvu:
@tiwillia Please take a look :)

node-util/sbin/oo-admin-cartridge
```diff
@@ -124,7 +124,7 @@ parser  = OptionParser.new do |opts|
   opts.on('-o', '--cartridge_vendor VENDOR', 'Cartridge vendor of cartridge to erase. default: redhat') { |o| options.cartridge_vendor = o }
   opts.on('-n', '--name NAME', 'Name of cartridge to erase') { |o| options.name = o }
   opts.on('-v', '--version VERSION', 'Version of packaged software to erase') { |o| options.version = o }
-  opts.on('-c', '--cartridge_version VERSION', 'Version of cartridge to erase') { |o| options.cartridge_version = o }
+  opts.on('-c', '--cartridge_version VERSION', 'Version of cartridge to erase (Note: All software versions of given cartridge version will be erased.)') { |o| options.cartridge_version = o }
```
tiwillia: Shouldn't this not be on the `-v` option?

Also, since all software versions will be deleted, would it be better to modify the `oo-admin-cartridge -c erase` logic to not require both the `-c` and `-v` options?

dinhxuanvu: I can modify the info to be on the 'v' option but the erase method requires both -c and -v option in order to retrieve the correct cartridge. I would rather not to mess with it as it's working fine. Also, requiring to have both version and cart version may help to avoid some weird corner cases when somehow they have same different versions with the same cartridge version.

dinhxuanvu:
@tiwillia PTAL :D

tiwillia:
Looks good by me, and no reason to test a man page change.

[merge] plz!

PR #6399, opened 2016-07-11
Update resource_limits.conf

thrasher-redhat:
Bug 1196783
https://bugzilla.redhat.com/show_bug.cgi?id=1196783

When the memory_limit_in_bytes variable in resource_limits.conf is updated, the
OPENSHIFT_GEAR_MEMORY_MB env variable does not get updated for existing gears.
Now there is an additional note for users to run a workaround to update that
variable for existing gears.

### DIFF
--- a/node/conf/resource_limits.conf
+++ b/node/conf/resource_limits.conf
@@ -7,6 +7,12 @@
 #       oo-pam-enable --with-all-containers
 #       oo-admin-ctl-tc restart
 #
+# NOTE: If you change the memory_limit_in_bytes variable, run this command
+#       to update existing gears. Replace 512 with the new value in megabytes.
+#       Note the change in units from bytes to megabytes - memory_limit_in_bytes
+#       expects the value in bytes, where OPENSHIFT_GEAR_MEMORY_MB expects the
+#       value in megabytes.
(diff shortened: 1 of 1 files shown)


thrasher-redhat:
@tiwillia 

For review at your leisure.  Debating whether I should put a comment directly on the memory_limit_in_bytes line.

dinhxuanvu:
LGTM. Just a simple doc change. @abhgupta Would you mind reviewing/merging this since Tim and Miciah are both in PTO.

abhgupta:
@thrasher-redhat I agree that placing the comment right above the memory_limit_in_bytes line would make more sense.

Otherwise, LGTM

thrasher-redhat:
@abhgupta Updated the comment to appear both at the top (along with the other instructions for what to do when updating the file) and just above the variable itself.

abhgupta:
[merge] lgtm

Miciah:
Dependency problems again:
```
Gem::InstallError: json requires Ruby version ~> 2.0.
An error occurred while installing json (2.0.1), and Bundler cannot continue.
Make sure that `gem install json -v '2.0.1'` succeeds before bundling.
```

abhgupta:
re-[merge] the flake was due to a gem dependency issue and was resolved in a separate PR

PR #6403, opened 2016-07-25
Allow maximum allowed user defined env variables to be configured

tiwillia:
Bug 1266239
https://bugzilla.redhat.com/show_bug.cgi?id=1266239

Adds the `USER_VARIABLE_MAX_COUNT` option to the node configuration file. This option controls the maximum number of user environment variables allowed on the node.

### DIFF
--- a/node/lib/openshift-origin-node/model/application_container_ext/environment.rb
+++ b/node/lib/openshift-origin-node/model/application_container_ext/environment.rb
@@ -10,7 +10,6 @@ module OpenShift
         include Kerberos
         include SecureShell
 
-        USER_VARIABLE_MAX_COUNT      = 50
         USER_VARIABLE_NAME_MAX_SIZE  = 128
         USER_VARIABLE_VALUE_MAX_SIZE = 512
         RESERVED_VARIABLE_NAMES      = %w(OPENSHIFT_PRIMARY_CARTRIDGE_DIR OPENSHIFT_NAMESPACE PATH IFS USER SHELL HOSTNAME LOGNAME)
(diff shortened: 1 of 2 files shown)


tiwillia:
[test]

@thrasher-redhat can you review please?

thrasher-redhat:
LGTM

tiwillia:
re-[test] please.

tiwillia:
@thrasher-redhat can you give this one more quick look? Made a small change so that the default '50' would work.

thrasher-redhat:
LGTM.  Default 50 sounds solid.

node/lib/openshift-origin-node/model/application_container_ext/environment.rb
```diff
@@ -273,13 +272,18 @@ def validate_ssh_keys(ssh_keys)
 
         # Add user environment variable(s)
         def user_var_add(variables, gears = [])
+          config = ::OpenShift::Config.new
```
Miciah: We should add `require 'openshift-origin-common'` if we're going to use `OpenShift::Config`.

node/lib/openshift-origin-node/model/application_container_ext/environment.rb
```diff
@@ -273,13 +272,18 @@ def validate_ssh_keys(ssh_keys)
...
+          # Default to 50 maximum allowed user variables
+          user_variable_max = config.get('USER_VARIABLE_MAX_COUNT') || "50"
+          user_variable_max_count = user_variable_max.to_i
```
Miciah: Alternatively, `user_variable_max_count = config.get('USER_VARIABLE_MAX_COUNT', '50').to_i`.

node/lib/openshift-origin-node/model/application_container_ext/environment.rb
tiwillia: Why? Other modules under ApplicationContainerExt that use `OpenShift::Config` do not `require 'openshift-origin-common'`. I believe this is because `../application_container.rb` contains `require 'openshift-origin-common'`

Miciah: Generally, if a file uses a library, the file should include the library, in case it is ever loaded without first loading another file that loads the library, or in case that latter file is changed in the future not to require the library.  However, I don't feel strongly enough about it to argue or hold up the PR.

Miciah:

tiwillia:
Using `config.get('VALUE', default)`, tests seem to be unable to determine the default value. I can confirm that in actual use, the default is properly established when `'VALUE'` does not exist. However the default is not established in tests.

@Miciah, I've modified the line to match what other application_container_ext modules are doing, like so:
```
user_variable_max_count = (config.get('USER_VARIABLE_MAX_COUNT') || "50" ).to_i
```

This works in tests and in the real world.

tiwillia:
re-[test] please

PR #6405, opened 2016-07-25
routing-daemon: Resynchronize with load-balancer

Miciah:
Resynchronize the routing daemon with the load balancer when the daemon attempts to perform on operation on an unrecognized pool or pool member.

The intention of this commit is to improve the daemon's reliability when something other than the current instance of the daemon changes the load balancer's state, for example the system administrator or clustered routing daemons.

`AsyncLoadBalancerController::Pool`, `BatchedLoadBalancerController::Pool`, `LoadBalancerController::Pool`, `SimpleLoadBalancerController::Pool`: Add `synch` method.

`AsyncLoadBalancerController`, `BatchedLoadBalancerController`, `LoadBalancerController`, `SimpleLoadBalancerController`: Add `synch_pools` method.

`RoutingDaemon`: Add `with_pool` method, which checks whether a pool exists, resynchronizes the daemon with the load balancer if it is not found (using the new `synch_pools` controller method), optionally creates the pool if it is still not found, and invokes the supplied block if the pool is found.

`RoutingDaemon#create_application`: Take `pool_name` as a parameter rather than generating the value itself since callers will already have generated the value themselves.

`RoutingDaemon#delete_application`, `RoutingDaemon#add_endpoint`, `RoutingDaemon#remove_endpoint`, `RoutingDaemon#add_alias`, `RoutingDaemon#remove_alias`, `RoutingDaemon#add_ssl`, `RoutingDaemon#remove_ssl`: Use the daemon's new `with_pool` method in order to check whether the pool exists,  esynchronize the daemon with the load balancer if an operation is attempted on an unrecognized pool, and (in the case of add methods) create the pool if it does not exist.

`RoutingDaemon#remove_endpoint`: Use the controller's new `synch` method for pools to resynchronize the daemon with the load balancer if the operation is attempted on an unknown pool member.

### DIFF
--- a/routing-daemon/lib/openshift/routing/daemon.rb
+++ b/routing-daemon/lib/openshift/routing/daemon.rb
@@ -296,9 +296,26 @@ module OpenShift
       @monitor_path_format.gsub(/%./, '%a' => app_name, '%n' => namespace)
     end
 
-    def create_application app_name, namespace
+    def with_pool app_name, namespace, create_if_nil=true
       pool_name = generate_pool_name app_name, namespace
+      if @lb_controller.pools[pool_name].nil?
+        @logger.info "Pool #{pool_name} unrecognized; resynching pools..."
+        @lb_controller.synch_pools
+      end
 
+      if @lb_controller.pools[pool_name].nil?
+        if create_if_nil
+          create_application app_name, namespace, pool_name
+        else
+          @logger.info "Pool #{pool_name} not found; ignoring"
+          return
+        end
+      end
+
+      yield @lb_controller.pools[pool_name]
+    end
+
+    def create_application app_name, namespace, pool_name
       if @monitor_name_format && @monitor_name_format.match(/%a/) && @monitor_name_format.match(/%n/)
         monitor_name = generate_monitor_name app_name, namespace
         monitor_path = generate_monitor_path app_name, namespace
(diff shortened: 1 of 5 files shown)


routing-daemon/lib/openshift/routing/daemon.rb
```diff
@@ -333,65 +349,75 @@ def delete_application app_name, namespace
...
+          @logger.info "No member #{member} exists in pool #{pool.name}; ignoring"
+        end
       end
     end
 
     def add_alias app_name, namespace, alias_str
-      pool_name = generate_pool_name app_name, namespace
-      @logger.info "Adding new alias #{alias_str} to pool #{pool_name}"
-      @lb_controller.pools[pool_name].add_alias alias_str
+      with_pool app_name, namespace do |pool|
```
tiwillia: Should this be passing `create_if_nil=false`, since it was not creating the application previously?

tiwillia:
@Miciah Looks good to me sir, besides the one small comment I made.

I looked over the `RoutingDaemon` with these changes in mind too and I don't see it affecting anything there. Fairly confident that there aren't any other issues, although I havn't tested it myself.

routing-daemon/lib/openshift/routing/daemon.rb
Miciah: It makes sense to me that all the `add_` methods should as a rule create the pool if it does not exist.  Using the default of `create_if_nil=true` here does indeed change the behaviour, but I see no reason for `add_alias` to be an exception to that rule.  There is a risk if events are handled out of order (for example, by clustered routing-daemon instances, each picking event notifications off a queue), and in taking this approach, the result of handling operations out of order could be that we create an extra pool (for example, if we get a create-application event, an add-alias event, and a delete-application event and handle the last two out of order).  On the other hand, if we specified `create_if_nil=false`, then if corresponding create-application and add-alias events were swapped, the result would be that we would fail to create the pool and alias when we should have done so (the pool would be created by any subsequent add-public-endpoint event, but the alias would never be created).

Miciah:
Thanks!

tiwillia:
[merge] please!

PR #6406, opened 2016-07-26
[TO_STAGE] Making user provisioning configurable within the broker

abhgupta:
Currently, a lookup is made in the broker for an authenticated user.
If the user does not exist, a user document is created in mongo.
With this change, a user will be provisioned only if the configuration is set to false


a13m:
Can we merge this please @abhguptake ?

a13m:
Can we merge this please @abhgupta ?

abhgupta:
@a13m was working on it - just added another commit to make the error message configurable.

PR #6408, opened 2016-07-27
rsync_destination_container: Fix free space check

Miciah:
Use correct arithmetic when checking that the destination node has enough free space before copying a gear: the remaining space after the copying the gear should be _greater_ than 5% (not less), and the arithmetic must be performed using floating point arithmetic (not integer).

This commit is related https://github.com/openshift/origin-server/pull/6382.
## 

openshift-bot, please [test][extended:node]!
## 

@sallyom, @a13m, we are all implicated in this PR by our involvement in https://github.com/openshift/origin-server/pull/6382.

### DIFF
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -2416,7 +2416,7 @@ module OpenShift
 
       # check here to make sure addition of gear to destination_container
       # will not result in > 95% full destination_container
-      if (destination_avail_space - source_used_blocks.to_i)/destination_total_space > 0.05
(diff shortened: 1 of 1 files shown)


Miciah:
The concern in this PR and in https://github.com/openshift/origin-server/pull/6382 is that we do not want to move a gear to a node when doing so will increase the node's storage usage to 95% or more of available storage.  @abhgupta raises two concerns: (1) We should make this 95% number configurable, and (2) we may want to apply a similar restriction when the system administrator uses `oo-admin-move` without specifying a node, or when the platform is placing a new gear.  For (2), we would need to specify this as a criterion in the `find_available` method.  The difficulty there would be that we would not know the size of the gear in `find_available` (in the case of a new gear, the size cannot be known), so we would need to determine some reasonable fixed amount of space (for example, filter out nodes with less than 500 MiB of space available).

Miciah:

PR #6411, opened 2016-07-29
Use updated fakefs to ensure support for File.absolute_path

tiwillia:
The use of fakes-0.4.2 causes erroneous test failures due to the lack of `File.absolute_path` support. Updating to 0.5.\* allows tests to succeed without causing further issues.

In origin-server tests, fakefs-0.9 is used by default. This change will have no affect on origin-server tests, but will ensure that tests run in other deployments will be run properly.

@thrasher-redhat fyi

### DIFF
--- a/node/openshift-origin-node.gemspec
+++ b/node/openshift-origin-node.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |s|
   s.files       += %w(README.md Rakefile Gemfile rubygem-openshift-origin-node.spec openshift-origin-node.gemspec COPYRIGHT LICENSE)
   s.require_paths = ["lib"]
   s.add_dependency("json")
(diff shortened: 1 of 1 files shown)


tiwillia:
[test] just for fun's sake.

tiwillia:
@Miciah, can you review/merge this when you get a chance?

Miciah:
So in discussions on IRC, we thought the version of fakefs was too *new*, but it turns out it was too *old*; is that right?

tiwillia:
@Miciah that is correct. fakefs was too old and did not support some methods that have been implemented in `File`.

tiwillia:
Hold off on this please, don't merge yet.

tiwillia:
This isn't necessary, a newer version will always be used. Closing.

PR #6414, opened 2016-08-04
Fix pip cache directory permission denied

dinhxuanvu:
For Python 2.7, a new version of PIP (7.1.0) attemps to write into .cache
directory in gear's home directory which will cause permission denied error.
Users do not have permission to create files or folders in gear's home directory.
The .cache directory needs to be created during the installation process. As
a result, a piece of code is added to "install" script to create .cache dir
before pip install step happens to address the issue.

Bug 1361755
Link https://bugzilla.redhat.com/show_bug.cgi?id=1361755

### DIFF
--- a/cartridges/openshift-origin-cartridge-python/bin/install
+++ b/cartridges/openshift-origin-cartridge-python/bin/install
@@ -10,6 +10,7 @@ esac
 echo "$version" > ${OPENSHIFT_PYTHON_DIR}env/OPENSHIFT_PYTHON_VERSION
 
 mkdir -p ${OPENSHIFT_PYTHON_DIR}template
+mkdir $OPENSHIFT_HOMEDIR/.cache
 
 # Call the version specific install script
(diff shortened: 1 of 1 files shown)


dinhxuanvu:
@a13m PTAL. Thanks.

dinhxuanvu:
@tiwillia Please review and merge. Thanks :)

tiwillia:
I see this properly resolving issues for new gears, since they will call the install method for the python cartridge upon creation. But what about existing gears? Are existing gears affected by the updated PIP version?

dinhxuanvu:
The existing gears probably have an older version of PIP which will not have this issue. Only new deployment is affected by this.

tiwillia:
LGTM then. Lets [test][extended:cartridge]

dinhxuanvu:
Yay!... Please bless it with a merge tag @tiwillia ... Thanks

tiwillia:
:clap: :clap: :point_up: :clap: :clap: [merge]

PR #6415, opened 2016-08-10
Add new update instruction for different gear types conf

dinhxuanvu:
When the memory_limit_in_bytes variable in resource_limits.conf is updated, the
OPENSHIFT_GEAR_MEMORY_MB env variable does not get updated for existing gears.
Now there is an additional note for users to run a workaround to update that
variable for existing gears.

Bug 1196783
Link https://bugzilla.redhat.com/show_bug.cgi?id=1196783


dinhxuanvu:
@tiwillia PTAL.

tiwillia:
:+1: [merge]

PR #6416, opened 2016-08-10
Perform quota operations on gear, not application

tiwillia:
The get_quota and set_quota methods take a gear argument. An application
object will work as the argument due to supporting similar methods, such
as the UUID method. However, an application's uuid will differ from a
gear's uuid when predictable uuid's are used.

Bug 1365390
https://bugzilla.redhat.com/show_bug.cgi?id=1365390

### DIFF
--- a/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
+++ b/plugins/msg-broker/mcollective/lib/openshift/mcollective_application_container_proxy.rb
@@ -1800,8 +1800,7 @@ module OpenShift
     end
 
     def update_cluster(gear, options)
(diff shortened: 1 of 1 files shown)


tiwillia:
[test]

@vdinh PTAL

@sallyom FYI

tiwillia:
`update_cluster` should always be run from the first web_proxy gear as far as the controller is concerned. The first web_proxy gear should always be the head gear of the application. There shouldn't be any reason to use the gear's application to ensure the head gear is used.

dinhxuanvu:
LGTM

tiwillia:
[merge]

PR #6417, opened 2016-08-16
Revert dfac396 and enable routing notifications during moves

tiwillia:
The routing daemon was previously not receiving updated endpoint
notifications during gear moves. A previous commit, dfac396, enabled
endpoints to be published but the endpoints published were the same
endpoints as before the move.

This change reverts dfac396 and properly sends endpoint creation and
deletion events to the routing daemon during a gear move.

### DIFF
--- a/controller/app/models/application.rb
+++ b/controller/app/models/application.rb
@@ -1735,7 +1735,13 @@ class Application
         if gear
           pi = PortInterface.create_port_interface(gear, component_id, *command_item[:args])
           gear.port_interfaces.push(pi)
-          pi.publish_endpoint(self)
+          # If available, use the public external ip address provided by the cartridge command.
+          # It is possible that if the NOTIFY_ENDPOINT_CREATE command was generated during the
+          # move of a gear (as opposed to the creation of a gear), the address from the gear
+          # object may not yet be updated. We prefer to use the address specified in the command
+          # itself.
+          public_ip = command_item[:args][0]["external_address"] || gear.get_public_ip_address
+          pi.publish_endpoint(self, public_ip)
         end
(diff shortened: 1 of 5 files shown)


tiwillia:
@Miciah @sallyom PTAL

[test]

controller/app/models/application.rb
```diff
@@ -1735,7 +1735,8 @@ def process_commands(result_io, component_id=nil, gear=nil)
         if gear
           pi = PortInterface.create_port_interface(gear, component_id, *command_item[:args])
           gear.port_interfaces.push(pi)
-          pi.publish_endpoint(self)
+          public_ip = command_item[:args][0]["external_address"] || gear.get_public_ip_address
+          pi.publish_endpoint(self, public_ip)
```
tiwillia: @Miciah this change, as well as the change to port_interface.rb, are necessary due to the way the PortInterface class gets the external "public" ip address of the node the gear is on. When calling PortInterface.external_address, the gear proxy is used and will return the old ip address of the moved gear.

The correct external ip address is passed in the NOTIFY_ENDPOINT_CREATE arguments, so we use that instead if its available, similar to the way the external ip address is handled for NOTIFY_ENDPOINT_DELETE events.

Miciah: Can we add a short comment? Something like, "Note: If the NOTIFY_ENDPOINT_CREATE command was generated in the process of moving a gear (as opposed to creating a new gear), the address from the gear object may not yet be updated, so we prefer to use the address specified in the command itself if it is available."

Miciah:
Looks super!

tiwillia:
With a review and a successful test, lets [merge] please!

Miciah:
Yay!

PR #6422, opened 2016-09-20
Allow start_priority to determine whether a gear should be idled

a13m:
The gears which we set a special start priority on are also gears
which we do not want to idle.  This change allows the idler to
skip gears based on start priority, regardless of whether they
have been recently accessed.

### DIFF
--- a/node-util/sbin/oo-auto-idler
+++ b/node-util/sbin/oo-auto-idler
@@ -134,7 +134,7 @@ def populate_ignorelist()
 end
 
 # Idle all gears on the node unless the list only option has been passed
-def idle_gears(interval, list_only)
+def idle_gears(interval, list_only, start_priority)
   puts "Only listing idle gears" if list_only
   stale_gear_list = []
   most_recent_last_access = -1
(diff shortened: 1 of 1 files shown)


tiwillia:
[test] please.

tiwillia:
Failure appears to be unrelated:
```
          -----------------------------------------------------------
                      Begin Output From Site Check Javascripts Tests
          -----------------------------------------------------------

# cd openshift-test/site; bundle exec rake test:check:js

/opt/rh/ruby193/root/usr/bin/rake: No such file or directory - phantomjs --version
rake aborted!
PhantomJS returned non-zero exit status 127. Make sure phantomjs runs successfully on your system. You can test this by just running the `phantomjs` command which should give you a Javascript REPL.

Tasks: TOP => konacha:run
(See full trace by running task with --trace)

          -----------------------------------------------------------
                       End Output From Site Check Javascripts Tests
          -----------------------------------------------------------
```

tiwillia:
Looks like a flake to me:
```
# phantomjs --version
1.9.0
# echo $?
0
```

The first time I ran `phantomjs --version` it took almost 2 minutes to return. Returns almost immediately afterwards.

[test] again please.

tiwillia:
Test failure should be resolved, please re-[test]

tiwillia:
[merge] please!

PR #6423, opened 2016-09-21
routing-daemon: F5: Pin to API version 11.6.0

Miciah:
`F5IControlRestLoadBalancerModel#rest_request`: Specify iControl REST API version 11.6.0.

We have previously tested the routing-daemon successfully against F5 BIG-IP LTM 11.6 but have encountered problems with 12.1.  However, the F5 iControl REST API is supposed to provide backwards compatibility with older API versions if the API version is specified in the request.


tiwillia:
[merge] please!

tiwillia:
Test failure should be resolved, please re-[merge]

Issue #6390, opened 2016-06-06
PIP install broken with Python-2.7


dneyirp:
Hi 

I followed instructions from https://developers.openshift.com/languages/python/flask.html
to install a tutorial Flask app. A few days ago all worked but as of yesterday (05 May 2016) it is not working.

When I add "`Flask==0.10.1`" to "`requirements.txt`" file, commit and push with git I get the following error:
`remote: Collecting Flask==0.10.1 (from -r /var/lib/openshift/57550b1c0c1e666d9400012d/app-root/runtime/repo/requirements.txt (line 1))
remote:   The repository located at mirror1.ops.rhcloud.com is not a trusted or secure host and is being ignored. If this repository is available via HTTPS it is recommended to use HTTPS instead, otherwise you may silence this warning and allow it anyways with '--trusted-host mirror1.ops.rhcloud.com'.
remote:   Could not find a version that satisfies the requirement Flask==0.10.1 (from -r /var/lib/openshift/57550b1c0c1e666d9400012d/app-root/runtime/repo/requirements.txt (line 1)) (from versions: )
remote: No matching distribution found for Flask==0.10.1 (from -r /var/lib/openshift/57550b1c0c1e666d9400012d/app-root/runtime/repo/requirements.txt (line 1))`

As mentioned, this worked flawlessly a few days ago.

When I ssh into my app account and run:
`pip install Flask --no-cache-dir`
there are no problems installing Flask.

If I add `Flask==0.10.1` to `install_requires` in file `setup.py`. The installation works correctly with no problems.

Can someone help with this problem?

dneyirp:
Hi

Here is an update.

I have added this line to the top of the `requirements.txt` file:
`--index-url https://pypi.python.org/simple/`
which takes me a bit further in that I actually get the Flask package downloaded but the installation fails with this error:
`OSError: [Errno 13] Permission denied: '/var/lib/openshift/57550b1c0c1e666d9400012d/.cache'`

When I ssh into the app account and check with `pip list` Flask is still not installed.
I also do not see a `.cache` folder and neither can I create one.

So far only adding install dependencies to `setup.py` or ssh'ing and running pip without using the cache dir works.

dneyirp:
Here is a fix I managed to find using OpenShift actions hooks.
The key web links that helped are listed in the code listing below.

Detailed instructions:
* In your local repo, change to the folder `.openshift\action_hooks`
[...]

PR #6430, opened 2016-11-07
[TO_STAGE] Fix pip cache directory permission denied

tiwillia:
This should resolve issue #6390

### DIFF
--- a/cartridges/openshift-origin-cartridge-python/bin/install
+++ b/cartridges/openshift-origin-cartridge-python/bin/install
@@ -10,6 +10,7 @@ esac
 echo "$version" > ${OPENSHIFT_PYTHON_DIR}env/OPENSHIFT_PYTHON_VERSION
 
 mkdir -p ${OPENSHIFT_PYTHON_DIR}template
+mkdir $OPENSHIFT_HOMEDIR/.cache
 
 # Call the version specific install script
 exec ${OPENSHIFT_PYTHON_DIR}usr/versions/${version}/bin/install $version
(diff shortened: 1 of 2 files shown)


raditv:
👍  I hope it will be merged soon

thrasher-redhat:
[test]

tiwillia:
Added a python version bump from master.

Also added the previous version to the compatible-version list ONLY for stage to avoid a gear migration. We'll need to come up with a script to run against PROD that creates the $GEAR_HOME/.cache directory for each existing python-2.7 cartridge with the correct selinux contexts and unix permission bits.

tiwillia:
Lets [test] again with the most recent change please.

sferich888:
Does this not work? https://developers.openshift.com/languages/python/getting-started.html#step3

raditv:
@sferich888 it should work, thank you bro! I'll try it

tiwillia:
@sferich888 @raditv I don't understand what that blogpost is suggesting that differs from how we are already loading the virtual env. In a python 2.7 cartridge we use the below by default:
```
virtenv = os.environ['OPENSHIFT_PYTHON_DIR'] + '/virtenv/'
virtualenv = os.path.join(virtenv, 'bin/activate_this.py')
try:
    execfile(virtualenv, dict(__file__=virtualenv))
except IOError:
    pass
```

Its hard to read in the blogpost so maybe I'm reading incorrectly, but the blogpost suggests the exact same thing for python-2.7. The way it is loading the virtenv differently is suggested for python-3.3.

tiwillia:
I've confirmed with @sferich888 that no known workaround exists, we'll definitely need to get this change in.

wshearn:
[merge]

PR #6437, opened 2016-12-16
routing-daemon: Set ActiveMQ prefetch limit to 1

Miciah:
Set the ActiveMQ prefetch limit to 1 to prevent the ActiveMQ broker from sending messages faster than the routing-daemon handles them.

See <http://activemq.apache.org/what-is-the-prefetch-limit-for.html>.

This commit is related to bug 1227472.

https://bugzilla.redhat.com/show_bug.cgi?id=1227472

### DIFF
--- a/routing-daemon/lib/openshift/routing/daemon.rb
+++ b/routing-daemon/lib/openshift/routing/daemon.rb
@@ -197,6 +197,7 @@ module OpenShift
       subscription_hash = {
         'id' => @uuid,
         'ack' => 'client-individual',
+        'activemq.prefetchSize' => 1,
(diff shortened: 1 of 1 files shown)


thrasher-redhat:
> Consumers that are unable to cache prefetched messages must set their prefetch to 1. An example of such a consumer is one implemented using a scripting language like Ruby, say, that connects via STOMP.

Yup, thats us.  LGTM.

tiwillia:
LGTM :pizza:

Miciah:
Many thanks! openshift-bot, please [merge]!

PR #6441, opened 2016-12-16
routing-daemon: Improve logging of exceptions

Miciah:
`RoutingDaemon#listen`: When logging an exception, mention its class.  Also, when NACKing a message from ActiveMQ because an exception was raised while handling it, log the exception.  Finally, log ACKs (at debug level) in addition to NACKs (at info level).


thrasher-redhat:
LGTM

tiwillia:
LGTM :dancing_men:

Miciah:
Thank you both! openshift-bot, please [merge]!

PR #6442, opened 2016-12-20
routing-daemon: F5: Try harder to delete client-ssl profile

Miciah:
`F5IControlRestLoadBalancerModel#remove_ssl`: Try to delete the client-ssl profile even if deleting it from the virtual server fails; try to delete the key even if deleting the client-ssl fails; and try to delete the certificate even if deleting the key fails.

`LBModelExceptionCollector` New helper class to rescue, collect, and combine `LBModelException` exceptions.

F5 BIG-IP will not let us delete a resource that is in use (e.g., a client-ssl profile that is still associated with a virtual server), but this change will enable the routing-daemon to finish cleaning up a client-ssl profile or associated key or certificate that is already deleted from the virtual server.

This commit fixes bug 1406251.

https://bugzilla.redhat.com/show_bug.cgi?id=1406251

---

@thrasher-redhat

### DIFF
--- a/routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
+++ b/routing-daemon/lib/openshift/routing/models/f5-icontrol-rest.rb
@@ -358,17 +358,29 @@ module OpenShift
     end
 
     def remove_ssl pool_name, alias_str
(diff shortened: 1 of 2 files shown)


thrasher-redhat:
LGTM.  Always good to try harder.

Miciah:
Thanks!

PR #6445, opened 2017-01-06
investigate bug with stg quota

sallyom:
@tiwillia @thrasher-redhat @wshearn 
*  believe this fixes things... and no need to bump quota limits.  I am able to reproduce in devenv, and this change allows starts/stops/restarts. 
* I've tested with multi-node env/single node env/scalable/un-scalable.  With master I am able to reproduce this bug (in devenv) but with this change bug is fixed (hope it holds true in stg!). 
* I believe this was always an issue, but during original bugfix I only used multi-node devenv/scalable app so maybe it went undetected (by QE, too?).
* can we test in stg please?

### DIFF
--- a/node/lib/openshift-origin-node/model/v2_cart_model.rb
+++ b/node/lib/openshift-origin-node/model/v2_cart_model.rb
@@ -1385,10 +1385,20 @@ module OpenShift
 
             buffer << out if out.is_a?(String)
             buffer << err if err.is_a?(String)
-
-            raise ::OpenShift::Runtime::Utils::ShellExecutionException.new(
+            # We ignore 'Disk quota exceeded' msg to allow apps at quota to start/stop/restart.
+            # During moves, quota limits are bumped if necessary.  Here, no bump, just ignore.
+            # Since we aren't bumping quota here, no worries about user exploiting limits 
+            # indefinitely, users can only use up to current limit.  This is a convenience so
+            # nothing gets 'stuck'.
+            if %w[start stop restart].include?(action)
+              raise ::OpenShift::Runtime::Utils::ShellExecutionException.new(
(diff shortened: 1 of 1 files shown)


node/lib/openshift-origin-node/model/v2_cart_model.rb
```diff
@@ -1385,10 +1385,20 @@ def do_control_with_directory(action, options={})
...
-            raise ::OpenShift::Runtime::Utils::ShellExecutionException.new(
+            # We ignore 'Disk quota exceeded' msg to allow apps at quota to start/stop/restart.
+            # During moves, quota limits are bumped if necessary.  Here, no bump, just ignore.
+            # Since we aren't bumping quota here, no worries about user exploiting limits 
+            # indefinitely, users can only use up to current limit.  This is a convenience so
+            # nothing gets 'stuck'.
+            if %w[start stop restart].include?(action)
+              raise ::OpenShift::Runtime::Utils::ShellExecutionException.new(
+                        "CLIENT_ERROR: Failed to execute: 'control #{action}' for #{path}", rc, out, err
+                    ) if (rc != 0 && !out.include?("Disk quota exceeded") && !err.include?("Disk quota exceeded"))
```
tiwillia: The "Disk quota exceeded" string should be defined as a constant preferably. If we change the error message then this code will no longer work.

But this is all old and its extremely unlikely we'll change the error message... so this should be fine.

thrasher-redhat:
[test]

thrasher-redhat:
@sallyom These changes are still failing in stage.  They might still fix a corner case to the original PRs, but I'm not sure the stage failure is the same thing we've replicated in devenvs.

sallyom:
@thrasher-redhat @tiwillia can I get the logs from the most recent failures in stage? Thanks

sallyom:
resolved elsewhere, this PR no longer needed

PR #6446, opened 2017-02-23
Bug 1244126: Allowing 4096 char limit for env variable values

abhgupta:
@thrasher-redhat @tiwillia PTAL

Fixes Bug 1244126


tiwillia:
Do we want to update the nodes with the new default limit? Otherwise the api will pass validation, but unless the allowed size has been increased on the node, the environment variable will fail to be created. I don't think we have a value set up in node.conf for the maximum to be changed easily:

  https://github.com/openshift/origin-server/blob/master/node/lib/openshift-origin-node/model/application_container_ext/environment.rb#L14

This node test also needs to be updated if we increase the default:

  https://github.com/openshift/origin-server/blob/master/node/test/unit/application_container_test.rb#L366

PR #6449, opened 2017-07-21
Fix input/output deadlock condition

ironcladlou:
If the rate of input and output are very high, the message queue size
can trend towards zero. In this case, it's possible that the Input will
decide to evict from the queue, but by the time the eviction occurs the
output has drained the queue to zero, causing the eviction dequeue to
block forever. In this case, Input is now blocked trying to evict, and
Output is blocked waiting for Input.

Fix the deadlock by distinguishing between eviction of old messages and
dropping of new messages, and making those operations non-blocking.

Add a test which verifies the fix under these conditions.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1464125

### DIFF
--- a/logshifter/input.go
+++ b/logshifter/input.go
@@ -14,7 +14,9 @@ import (
 // be placed on the channel:
 //
 //   input.read => number of messages processed during a Read call
-//   input.drop => number of messages dropped from queue during a Read call
+//   input.queue => number of messages queued during a Read call
+//   input.drop => number of new messages dropped during a Read call
+//   input.evict => number of old messages evicted from the queue during a Read call
 //   input.read.duration => time taken to process a line from reader (micros)
 type Input struct {
 	bufferSize   int
@@ -40,6 +42,12 @@ func (input *Input) Read() *sync.WaitGroup {
 	return wg
 }
 
+func (input *Input) record(name string, val float64) {
+	if input.statsChannel != nil {
+		input.statsChannel <- Stat{name: name, value: val}
+	}
+}
+
 // Private synchronous portion of Read.
 func (input *Input) read() {
 	reader := bufio.NewReaderSize(input.reader, input.bufferSize)
@@ -47,10 +55,7 @@ func (input *Input) read() {
 	for {
 		line, _, err := reader.ReadLine()
 
-		var start time.Time
-		if input.statsChannel != nil {
-			start = time.Now()
-		}
+		start := time.Now()
 
 		if err != nil {
 			break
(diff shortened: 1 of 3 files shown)


logshifter/input.go
```diff
@@ -64,24 +67,39 @@ func (input *Input) read() {
...
 
-		if input.statsChannel != nil {
-			input.statsChannel <- Stat{name: "input.read", value: 1.0}
-		}
+		input.record("input.read", 1.0)
 
 		select {
 		case input.queue <- cp:
 			// queued
+			input.record("input.queue", 1.0)
```
Miciah: Could add the new "input.queue" key to the enumeration of stats in the comment for `struct Input`.  Likewise for "input.evict".

thrasher-redhat:
LGTM.  Currently having Miciah take a look as well.

logshifter/input.go
```diff
@@ -64,24 +67,39 @@ func (input *Input) read() {
...
+			// try to evict the oldest entry to make room
+			select {
+			case <-input.queue:
+				input.record("input.evict", 1.0)
+				// try again to queue
+				select {
+				case input.queue <- cp:
+					// queued
+					input.record("input.queue", 1.0)
+				default:
```
Miciah: Is there a way this could theoretically happen? AFAICT, `(input *Input) read()` is the only code that writes to the queue, it is running in at most one thread, the queue is not closed (in `(shift *Shifter) Start()`) until the thread says it is done (in `(input *Input) Read()`), the queue size is fixed, the number of items in queue was equal to at most the size of the queue, and we just dequeued an item, so it should not be possible that writing to the queue will block a second time.

Not that I'm opposed to guarding against theoretically impossible contingencies; I'm mostly curious.

logshifter/input.go
```diff
@@ -64,24 +67,39 @@ func (input *Input) read() {
...
+					// no room, drop it
+					input.record("input.drop", 1.0)
+				}
+			default:
+				// queue is already empty, try to queue
+				select {
+				case input.queue <- cp:
+					// queued
+					input.record("input.queue", 1.0)
+				default:
```
Miciah: Again, is there any way this could theoretically happen?

logshifter/input_test.go
```diff
@@ -0,0 +1,90 @@
...
+		queue:        queue,
+		bufferSize:   messageSize,
+		statsChannel: statsCh,
+	}
+	input.read()
+
+	// Shut down the stats collector.
+	close(stop)
+	statsWg.Wait()
+	t.Logf("stats: %#v", stats)
```
Miciah: Do I understand correctly that the way this test would fail would be by blocking so that control never reached this point? Might be worth adding a comment that if we get here without deadlock, the test has succeeded.

ironcladlou: You're correct; I added a comment as you suggested.

logshifter/input_test.go
```diff
@@ -0,0 +1,90 @@
...
+	var i int64 = 0
+	for ; i < msgCount; i++ {
+		data += strings.Repeat("0", msgLength-1) + "\n"
+	}
+
+	buffer := bytes.NewBufferString(data)
+
+	fmt.Printf("created %d byte test input (%d lines @ %d bytes each)\n", buffer.Len(), msgCount, msgLength)
+
+	return &testReader{buffer: buffer}
```
ironcladlou: You're right; I copied and gutted that structure from elsewhere, and in the gutting removed the reasons for the struct to exist. Simplified.

logshifter/input.go
ironcladlou: Done

logshifter/input.go
ironcladlou: I think we both came to the same conclusion here... I couldn't come up with a way it was possible given a single threaded reader, but felt I'd include coverage for the case anyway.

logshifter/input.go
ironcladlou: As above, I don't think it's possible, but wanted to cover it anyway.

logshifter/input_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// testReader is an io.Reader which will read from a predefined buffer and
+// returns an error when the buffer is empty.
+type testReader struct {
+	buffer *bytes.Buffer
+}
+
+func (r *testReader) Read(b []byte) (int, error) {
+	written, err := r.buffer.Read(b)
```
ironcladlou: Scrapped the whole struct per your following suggestion.

ironcladlou:
Addressed the great feedback from @Miciah (thanks!)

Miciah:
openshift-bot, do you remember how to [test][extended:cartridge,gear]?

thrasher-redhat:
Manual tests passed.  @ironcladlou can you squash these commits and then we should be good to go.

ironcladlou:
Squashed

abhgupta:
[merge] Reviewed by @thrasher-redhat and @Miciah

thrasher-redhat:
@abhgupta It looks like openshift-bot's v2 functionality has been removed.  Can you merge it manually?

abhgupta:
Tested by Rory/Tim locally.

